PR #40, opened 2022-08-11
[FINISH TITLE]

### LINKED ISSUE #33: Create game layer to track progress
## Issue Description

Create a game layer that enables things like progression tracking by converting a scenario and its scenes into a graph representation

See minutes from 3/8/22 for more details: https://docs.google.com/document/d/1tK8Goe7VbGD-C0zOAQzgCRt4Zg3RCA6aZ1Z6QVdvDkQ/edit?pli=1#bookmark=id.9qj4xihvb8lv

## User Story

"As a user, I want the app to track my progress as I play a scenario"

## MVP

This one's a little tricky, but something like a graph representation of a scene and backend functionality to calculate things like how many scenes are left until the end scene.

## Tasks

Describe the tasks in this story.

- [x] Create endpoint that returns a graph representation of the scenario
- [x] Create Graph class
- [x] Create method to track the distance between any two scenes
- [x] Create method to track the progress of a player

Ray-F:
## Describe the issue

A clear and concise description of what the issue is.

## Describe the solution

A clear and concise description of what the solution is.

## Risk

Potential risks that this PR might brings

## Definition of Done

- [ ] Code peer-reviewed
- [ ] Acceptance criteria met

## Reviewed By

Closes #33


backend/src/models/ScenarioGraph.js
```diff
@@ -0,0 +1,123 @@
...
+
+    constructor() {}
+
+    // Getters
+
+    get length() { return this.queue.length }
+
+    get isEmpty() { return this.queue.length === 0 ? true : false}
+
+    get queue() { return this.queue }
```
Ray-F, line 113: You should remove this method since you never want to expose the underlying data structure behind an implementation

backend/src/models/ScenarioGraph.js
```diff
@@ -0,0 +1,123 @@
...
+                name: scene.name,
+                endScene: leaf,
+                linkedScenes: children
+            }
+        })
+    }
+
+    // Getters
+
+    get graph() { return this.graph }
```
Ray-F, line 39: See other comment - you should not expose underlying structure

UPDATED: I see you're returning this from the endpoint, so maybe add a comment to remove this later once confirmed what that endpoint will return?

backend/src/models/ScenarioGraph.js
```diff
@@ -0,0 +1,123 @@
...
+
+            // Iterate over each component of a scene
+            scene.components.forEach((component) => {
+                
+                // Check if the scene has any links to other scenes
+                if (component.type === 'BUTTON') {
+                    children.push(component.nextScene);
+                    leaf = false;    // Scene is not a leaf as it has a child.
+                }
+            })
```
Ray-F, line 26: Some JS syntactical sugar with functional programming if you want to use it to simplify your code :P 

```
                  
const children = scene.components
                  .filter((it) => it.type === 'BUTTON'))
                  .map((it) => it.nextScene));
// If the current scene has no children, then by definition it is a leaf node
const leaf = children.length === 0;

File: backend/src/routes/api/scene.js [hunk 4]
```diff
@@ -27,10 +29,27 @@ router.get("/full/:sceneId", async (req, res) => {
...
+// Get graph
+router.get("/graph", async (req, res) => {
+  const scenes = await retrieveSceneList(req.params.scenarioId);
+  let fullScenes = [];
+  
+  // Get FULL scene data - should really change this to run in parallel for better runtime
+  for await (const scene of scenes) {
+    const fullScene = await retrieveScene(scene._id);
+    fullScenes.push(fullScene);
+  }
```
Ray-F, line 44: You can parallelize async requests using `Promise.all()` - might be a good interim feature to speed up this process (which I think is worth it for this). Basically, something like...

```
const promises = scenes.map((it) => retrieveScene(it._id));
const fullScenes = await Promise.all(promises);
```

However, this still means we will make `n` requests to the mongoDB which isn't too great. See below comment for future improvement.

In the future, it'll be better if you could actually just pull all the info needed from the scenes with one request to MongoDB. @nickykay98 @Harshal-D this could be made into an improvement task (something like: create method `retrieveSceneListDetailed()` in `sceneDao.js` to retrieve all scenes and their details by providing a scenario ID).

backend/src/routes/api/scene.js
```diff
@@ -27,10 +29,27 @@ router.get("/full/:sceneId", async (req, res) => {
...
+
+  const scenarioGraph = new ScenarioGraph(fullScenes);
+  const distance = scenarioGraph.distanceBetween("6295a71dc02aa66b701339b2", "62f4805867e5e917c0ef095c");
+  console.log(distance);
```
Ray-F, line 48: I assume this is for testing? 

Maybe now a good time to chat with @Harshal-D and @nickykay98 about how this endpoint should be used, and this will link into the idea of "bad ends" and "successful end nodes". This will make the purpose of this endpoint a lot clearer.

Ray-F (commented):
Hey @DamonGreenhalgh, just added some comments and code improvements.. Overall great work though and it definitely is in a really great spot that can easily be built on for the future stats stories!

backend/src/models/ScenarioGraph.js
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Computes the distance between two scenes.
+     * @param startScene: string - the id of the starting scene
+     * @param endScene: string - the id of the end scene
+     * @returns depth: number - the distance from the startScene to the endScene.
+     *                          If endScene does not exist within the graph or 
+     *                          there is not directed path from startScene to
+     *                          endScene, return -1.
+     */
+    distanceBetween(startScene, endScene) {
```
Ray-F, line 52: Just a thought (no changes for now), but since the scene stays the same, we could actually generate a 2D distance map when the ScenarioGraph is constructed and save this data in like a 2D array (kinda like a 2D map).

Unsure about exact implementation, but constructor could build something similar to the `graph` field, but like...

`distances: Map<StartSceneId, EndSceneId => number>`

This means that provided this data model exists in the code (which in itself is related to caching in the backed – another story), we could get distances between any two scenes (which we will have to do every time the player moves to the next scene) without applying BFS n times.

backend/src/models/ScenarioGraph.js
```diff
@@ -0,0 +1,123 @@
...
+
+        let depth = 0;
+        let queue = new Queue();
+        let seen = [];
+        let found = false;
+
+        queue.enqueue(startScene);
+
+        // Apply BFS
+        while (!(queue.isEmpty || found)) {
```
Ray-F, line 66: NB: I'm assuming this algorithm works, not checking thoroughly 😛

Ray-F:
@DamonGreenhalgh also noticed that the linting annotations come up as well which kind of get in the way fo the code review. You can hide these by selecting the dots on the respective file and unticking "show annotations"
image

backend/src/models/ScenarioGraph.js
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/**
+ * Queue data structure for graph algorithms
+ */
+export class Queue {
+
+    queue = [];
+
+    constructor() {}
```
Ray-F, line 105: You can remove unused constructor (I think?) in JS

backend/src/models/ScenarioGraph.js
DamonGreenhalgh: Yes! This is a much better solution I wish I had thought of. Definitely would improve runtime with O(1) indexing at the cost of O(V^2) space. Building the distance matrix would be O(V^3) with something like Floyd's algorithm, but would only happen when we initialize the graph object.

Looking back at my original implementation looks to be around O(V^3) (should have used something like Dijkstra), this would be a huge improvement. This would also help with server overhead if multiple people are playing scenarios.

nickykay98:
@DamonGreenhalgh Looks like this was closed without merging into main?

DamonGreenhalgh:
That was my bad, I was so used to merging into a dev branch that I had just assumed this would merge into the wrong branch.

nickykay98:
Is it all good then? Because even I don't have the option to reopen it or anything 😅

DamonGreenhalgh:
Yeah, I opened a new PR but it is not linked to the issue. Just failing one test, but not sure how to solve as on my end it is fine.

nickykay98:
Yeah I was about to say you need to make a new PR (closed = "rejected PR, not merged" apparently)

I've linked the new one to the issue on my end
image

Issue #83, opened 2023-05-10
Delete 2 or more scenarios by using shift + click shortcut


Ryuna001:
- Delete 2 or more scenarios by using the combination shift + click (Bypass delete scenario confirmation pop-up)

Issue #82, opened 2023-05-10
Add keyboard and mouse shortcuts to delete scenarios


Ryuna001:
**User Story**

**Description**

Include keyboard and mouse combination shortcuts to improve efficiency in the scenario page.

- Delete a scenario by using the combination ctrl + click (Bypass delete scenario confirmation pop-up)
- Delete 2 or more scenarios by using the combination shift + click (Bypass delete scenario confirmation pop-up)

Issue #69, opened 2023-04-25
Add right click functionality to the scenario page


Ryuna001:
**User Story**

- [ ] #65 

**Description**

- Add a mouse event which displays all the other options also available to the right of the screen
- clicking on the option should perform the same tasks as clicking the button would

Issue #68, opened 2023-04-25
Add right click functionality on the home page


djos192:
**User Story**

- [ ] #64 

**Description**

- Add a mouse event which displays all the other options also available to the right of the screen
- clicking on the option should perform the same tasks as clicking the button would

PR #114, opened 2023-08-03
Vps 69/right click scene scenario

quirked-up:
## Describe the issue

As in #69 and #68 , we wish to be able to right click and access actions on the scenarios and scenes. This is because having to click on the scen(e/ario) and then click on the toolbar button takes a while.  

## Describe the solution
I added a right click menu as specified to the main scenario selection page, as well as the scene selection page.
image

On the technical side, I used material ui's Menu and MenuItem components to create a custom context menu (right-click menu), I created a new component ContextMenu that attaches the correct attributes onto the Menu, and should be used as defined in the ContextMenu.js' example. (Unfortunately I couldn't figure out a way to not have to define the handleContextMenu function in each component that uses ContextMenu, as it has to be defined on a div above the ContextMenu to properly handle subsequent right clicks)
I then added these to the scene and scenario selection page. I had to copy over a couple of functions form the ToolBar component into the ScenarioSelectionPage in order to pass them into the MenuItems.

## Risk
I'm not entirely certain as to why some of the classes changed on my snapshot update (penultimate commit), I think its from previous PRs that had broken testing but I'm not 100% sure, it'd be good to check that this isn't due to any bug, though everything looks normal from my own testing.

If/when we do multiple selection, as in #82 and #83, the functions on the MenuItem's will need to be revised to account for not having just 1 selected, and the current behaviour of right clicking first as if you left clicked first will need to be adjusted so as to not undo the current selection.

## Definition of Done

- [ ] Code peer-reviewed
- [ ] Acceptance criteria met

## Reviewed By


quirked-up:
Frontend prettier & unit tests are working on my machine, I think the current ones on main are broken due to the previous PR #102 .

quirked-up:
Hmm, to be thorough I think I need to add a ContextMenu.test.js

djos192 (approved):

Issue #138, opened 2024-03-18
[VPS-138] The 'Create New Scene' button has a giant **+** in the middle of it


JordanBlenn:
**Description:**

**AS A:** User
**I WANT:** The create scene button to have a + in the middle of it
**SO THAT:** It looks more like a create scene button

**AC's:**
1) When editing a scene, the create scene button has a giant **+** in the centre of it, with a max height of 80% the height of the component and a max width of 80% the width of the component.

**Technical Discussion:**

**Other Notes:**
Originally we also had AC 2 as part of this user story, but after discussion it has been deemed out of scope and will be fixed in a future issue. 
2) The **+** dynamically resizes as the window size changes, always restricted 80% width/height of the minimum of the height or width of the component.
To be fixed in #146

wjin-lee:
We should maybe change the second acceptance criteria?
Currently, the size of the scene elements (boxes) are relative to screen width at 20%. (I.e. it will ALWAYS show ~5). See case B in screenshot.
We could instead make this a fixed length or defined in `rem` and decoupling it from screen size. See case A.
![image](https://github.com/UoaWDCC/VPS/assets/100455176/c432900c-a145-4bfb-9414-f98936192665)

JordanBlenn:
[...]

PR #144, opened 2024-03-26
Added the plus sign

xche529:
## Describe the issue

The create scene button need to have a + in the middle of it

## Describe the solution

Added two boxes to form a +

## Risk

Potential styling issue


wjin-lee:
I don't think this technically meets AC 2. But I also think AC2 may be flawed.
Commented some thoughts on #138  and waiting for @JordanBlenn's comment on this as well

JordanBlenn (approved):
Looks good to me. Will address the dynamic resizing issue with other containers in #146, and AC's have been updated to reflect this.

Issue #154, opened 2024-04-14
[VPS-154](2)(0.5) Uploading a CSV file creates appropriate groups and assigns roles (No UI)


JordanBlenn:
**Description:**

**AS A:** Staff member creating scenarios
**I WANT:** to be able to upload a csv file containing group information
**SO THAT:** Groups are created with roles that people can play through

**AC's:**
1) A .csv file of a specific and exact format [email, name, group number, role], if uploaded to the system:
- Creates Group entries in the database 

**Technical Discussion:**
- The group entry/object contains a list of users that defines their roles (including name and email)

wjin-lee:
During development - just make a dummy `<input>` with the type of file. 
See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file

JordanBlenn:
Does the current implementation of CSV uploading allow a new CSV to be uploaded overriding the previous one? If so (and I think it should in case the wrong file is uploaded), it is important that roles are stored in such a way that if the name of a role is changed it doesn't impact other features we have yet to implement (such as assigning roles to specific scenes etc). We possibly could force people to always use the same number of roles (i.e. if they initially upload a CSV file with 3 roles they have to always use 3 roles no matter what for this scenario) if we warn them when the first upload a CSV. Most of this implementation can be done in a future story but this is something to keep in mind for overall design considering some future stories which rely on the list of roles determined by the CSV file.

Alternatively, we have the user input roles separately and store them there, and uploading a CSV file requires the same roles as headings or it doesn't work? 

Keen to hear your thoughts @Kot6603 @laibatool792 @wjin-lee

JordanBlenn:
> Does the current implementation of CSV uploading allow a new CSV to be uploaded overriding the previous one? If so (and I think it should in case the wrong file is uploaded), it is important that roles are stored in such a way that if the name of a role is changed it doesn't impact other features we have yet to implement (such as assigning roles to specific scenes etc). We possibly could force people to always use the same number of roles (i.e. if they initially upload a CSV file with 3 roles they have to always use 3 roles no matter what for this scenario) if we warn them when the first upload a CSV. Most of this implementation can be done in a future story but this is something to keep in mind for overall design considering some future stories which rely on the list of roles determined by the CSV file.
> 
> Alternatively, we have the user input roles separately and store them there, and uploading a CSV file requires the same roles as headings or it doesn't work?
> 
[...]

PR #177, opened 2024-05-01
Vps 151/notes-backend

xche529:
## Describe the issue

There's no backend structure for note feature

## Describe the solution

Created note schema in db and basic backend api

## Risk

Potential bug with db operations which may delete other information

## Definition of Done

- [ ] Code peer-reviewed
- [ ] Acceptance criteria met

## Reviewed By

### DIFF
--- a/backend/src/db/daos/noteDao.js
+++ b/backend/src/db/daos/noteDao.js
@@ -8,20 +8,28 @@ import Group from "../models/group";
  * @param {String} role role of the note
  * @returns
  */
-
 const createNote = async (groupId, title, role) => {
   const dbNote = new Note({ title: title, role: role });
   await dbNote.save();
   const updateQuery = {};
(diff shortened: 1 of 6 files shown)


wjin-lee (commented):
Nice job on the PR - really impressed by your quality of work!
Mainly just nits on the code front.

We will have to be careful here with #154 - notes field within groups schema will be changing a little bit to make querying for specific roles easier through Mongo.

Have a chat with Koutaro before merging this one to work out how notes is going to be initialised. If we do indeed go with a separate notes collection like you have here, there will need to be additional logic to initialise these during group creation.

If we keep it just as a string field within groups, we could still hold on to this in case Natalie wants to have multiple notes (which is likely imo).

Would be great if you could lmk of the discussion between you two before starting on your notes implementation (whether you use this implementation or Koutaro's group schema directly).

PR #180, opened 2024-05-13
VPS-164-UI for invalid scene permissions page

jacobmathew105:
Issue: A standalone page is required to redirect to when the user tries to access a scene that their role cannot access

Solution: A page is implemented with a back button to the scene scenario selection page

## Describe the issue
A standalone page is required to redirect to when the user tries to access a scene that their role cannot access

## Describe the solution

 A page is implemented with a back button to the scene scenario selection page
image

## Risk

Should be incorporated with the actual roles instead of dummy data

## Definition of Done

- [ ] Code peer-reviewed

## Reviewed By
Jordan and Jin


JordanBlenn (changes requested):
In the top line of each of your files `BackButton.jsx` and `InvalidRolePage.jsx` you've included the line `import React from "react"` but this line isn't actually needed. You should just be able to change it on your local build, commit and push to fix this (you've already created a pull request so you don't need to make another one, the changes still apply to this branch which still is being pulled into master if that makes sense).

JordanBlenn (approved):
Looks good to me :)

frontend/src/containers/pages/InvalidRolePage.jsx
```diff
@@ -0,0 +1,34 @@
+import BackButton from "../../components/BackButton";
+
+function InvalidRolePage() {
+  const currentUserRole = "Doctor";
+  const rolesWithAccess = ["Nurse", "Patient"];
+
+  const containerStyle = {
+    fontFamily: "Sitka Text, sans-serif",
```
wjin-lee, line 8: This is a custom font and you didn't import it in the rest of the application. What is its purpose here?

frontend/src/components/BackButton.jsx
```diff
@@ -0,0 +1,30 @@
...
+    display: "flex",
+    justifyContent: "center",
+    alignItems: "center",
+  };
+
+  const buttonStyle = {
+    padding: "10px 20px",
+    fontSize: "16px",
+    borderRadius: "5px",
+    backgroundColor: "#0d37de",
```
wjin-lee, line 14: This background colour does not match the application theme. **We should change it.** 

On Windows and MacOS, you can get extensions/applications that can get colour hexcodes from anywhere in your screen - checkout PowerToys for Windows.

frontend/src/components/BackButton.jsx
```diff
@@ -0,0 +1,30 @@
+import { Link } from "react-router-dom";
+
+const BackButton = () => {
```
wjin-lee, line 3: Could we name this better?

`BackButton` is ambiguous and potentially even misleading.
`BackToScenarioSelection` or similar would be more appropriate given what this does. (Redirects to root)

frontend/src/components/BackButton.jsx
```diff
@@ -0,0 +1,30 @@
...
+    backgroundColor: "#0d37de",
+    color: "#fff",
+    textDecoration: "none",
+  };
+
+  return (
+    <div style={containerStyle}>
+      <Link to="/" style={{ textDecoration: "none" }}>
+        <button type="button" style={buttonStyle}>
+          Go back
```
wjin-lee, line 23: Could we potentially give this a more descriptive label?

In the wireframes "Back to Scenarios" were used - this has the additional benefit of clearly telling the user where they will be navigated.

frontend/src/containers/pages/InvalidRolePage.jsx
```diff
@@ -0,0 +1,34 @@
...
+  };
+
+  return (
+    <div style={containerStyle}>
+      <h1 style={textMargin}>
+        Someone else is playing through this section of the scenario!
+      </h1>
+      <p style={textMargin}>Please wait for your role: {currentUserRole}</p>
+      <BackButton />
+      <p style={textMargin}>Roles with access: {rolesWithAccess.join(", ")}</p>
```
wjin-lee, line 29: Hmm, in the wireframe, this text is near the bottom. As you seem to have done this with flex boxes (nice!), think about how you can use multiple divs to group and position in a 'responsive' way.

image

frontend/src/containers/App.jsx
```diff
@@ -22,6 +23,15 @@ export default function App() {
         <AuthenticationContextProvider>
           <BrowserRouter>
             <Switch>
+              <ProtectedRoute
+                exact
+                path="/play/invalid-role"
+                component={InvalidRolePage}
+              >
+                <PlayingScenarioContextProvider>
```
wjin-lee, line 31: This has been replaced - and even for current PR, it could be argued that this is dead code.

Let's remove it :)

wjin-lee (changes requested):

frontend/src/containers/pages/InvalidRolePage.jsx
jacobmathew105: image

updated, looks like that now.

frontend/src/containers/App.jsx
jacobmathew105: Removed the playingscenariocontext provider from code

wjin-lee (approved):
Overall lgtm!

Note that [frontend/src/components/BackButton.jsx](https://github.com/UoaWDCC/VPS/pull/180/files#diff-518deb14b3187189d6aa0294772938bda1e59ee681480b9b2dc19d8f8f3ecea4) is deadcode. removal is required.

PR #184, opened 2024-05-16
Vps 174/single scenario last scene

harbassan:
## Describe the issue

A user's last visited scene doesn't save, so they cant play through a scenario across different times or across different devices.

## Describe the solution

- Add a new field on the user db schema that stores the path for multiple scenarios
- Add a route for updating this path as the user moves across scenes
- Change the user/scenario/data route to give the current scene for users as well as groups 

## Risk

Shouldn't bring any risks, as the existing route is only used in one place which has been updated accordingly.

## Definition of Done

- [x] Code peer-reviewed
- [x] Continuous Integration build passing
- [x] Acceptance criteria met

## Reviewed By

### DIFF
--- a/backend/.eslintrc.json
+++ b/backend/.eslintrc.json
@@ -26,6 +26,7 @@
     "jest/prefer-to-have-length": "warn",
     "jest/valid-expect": "error",
     "no-unused-vars": "warn",
-    "func-names": "off"
+    "func-names": "off",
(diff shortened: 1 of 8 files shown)


wjin-lee (approved):
lgtm! 

If we have time, we should clean up and refactor the redundant path updates being sent on last scene, but that is out-of-scope for now (and potentially this year? not sure).

PR #205, opened 2024-07-22
Vps 12/note backend authentication

xche529:
## Describe the issue

The backend API for note is not protected 

## Describe the solution

Add auth protect to note api and role checking logic to database opreations.

## Risk

Potential misuse of AuthenticationContext

## Definition of Done

- [ ] Code peer-reviewed
- [ ] Acceptance criteria met

## Reviewed By

### DIFF
--- a/backend/src/db/daos/noteDao.js
+++ b/backend/src/db/daos/noteDao.js
@@ -55,6 +55,7 @@ const deleteNote = async (noteId, groupId, email) => {
   await Group.updateOne({ _id: groupId }, updateQuery);
   //  delete note from note collection
   await note.delete();
+  return null;
 };
 
 /**
@@ -78,14 +79,12 @@ const updateNote = async (noteId, updatedNote, groupId, email) => {
   }
   const note = await Note.findById(noteId);
   if (note.role !== role) {
-    return null;
+    return;
   }
   note.title = updatedNote.title;
   note.text = updatedNote.text;
   note.date = updatedNote.date;
   await note.save();
-
-  return note;
 };
 
 /**
--- a/backend/src/db/daos/noteDao.js
+++ b/backend/src/db/daos/noteDao.js
@@ -46,7 +46,7 @@ const deleteNote = async (noteId, groupId, email) => {
   if (note.role !== role) {
     return null;
   }
-  
+
   const updateQuery = {
     $pull: { [`notes.${note.role}`]: noteId },
   };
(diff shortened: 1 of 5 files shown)


backend/src/db/daos/noteDao.js
```diff
@@ -8,7 +8,17 @@ import Group from "../models/group";
  * @param {String} role role of the note
  * @returns
  */
-const createNote = async (groupId, title, role, text = "") => {
+const createNote = async (groupId, title, email, text = "") => {
+  const dbGroup = await Group.findById(groupId);
```
wjin-lee, line 12: nit: the auth check could be abstracted out to comply with DRY.

E.g. method that retrieves the requesting user's role could be a potential path.

frontend/src/components/Note.jsx
```diff
@@ -23,7 +24,11 @@ export default function Note({ role, id, group, user, refetchGroup }) {
   };
 
   async function loadNote() {
-    const noteData = await usePost("/api/note/retrieve", { noteId: id });
+    const noteData = await usePost(
+      "/api/note/retrieve",
+      { noteId: id },
+      getUserIdToken
```
wjin-lee, line 30: Unsafe usage of `getUserIdToken` - have a read of https://wdcc-vps.atlassian.net/wiki/spaces/W/pages/14417933/Auth+Implementation+Guide

notably the user stateful value and the other error and loading cases.

frontend/src/components/Note.jsx
```diff
@@ -71,8 +82,8 @@ export default function Note({ role, id, group, user, refetchGroup }) {
     try {
       await saveNote();
```
wjin-lee, line 83: Unsafe usage of `getUserIdToken` in `saveNote` and `loadNote` - see above comment

frontend/src/components/NotesDisplayCard.jsx
```diff
@@ -38,11 +41,15 @@ export default function NotesDisplayCard({ group, user, handleClose }) {
...
-      role: userRole,
-    });
+    await usePost(
+      "/api/note/",
+      {
+        groupId: group._id,
+        title: "New Note",
+        email: user.email,
+      },
+      getUserIdToken
```
wjin-lee, line 51: Unsafe use of getUserIdToken - see above comment

wjin-lee (changes requested):
A few unsafe getUserIdToken usages - looking good otherwise!

backend/src/db/daos/noteDao.js
harbassan: I've made and am using a function for this already, you could just move it to the groupDao file and then use it from there.

https://github.com/UoaWDCC/VPS/blob/58baaf46e9eb3aeacfc9e7b7ecd5316cfc8ec44b/backend/src/routes/api/navigate/group.js#L38

xche529: Cool, I will try to write a method to check if the user is within a group because it's been used alot

backend/src/db/daos/noteDao.js
```diff
@@ -1,14 +1,36 @@
...
 import Group from "../models/group";
 
+/**
+ * Checks if a user is in a group
+ * @param {String} groupId group ID
+ * @param {String} email email of the user
+ * @returns role of the user in the group
+ * @returns null if user is not in group
+ */
+const checkRole = async (groupId, email) => {
```
wjin-lee, line 11: nit: more appropriate name would be something along the lines of "getRole" or "findRole".

frontend/src/components/Note.jsx
wjin-lee: This is still unsafe right? What if `getUserIdToken` returns a null result? What then? What if it's simply authenticating again and it will eventually return a value? What if the authentication fails for whatever reason?

wjin-lee (changes requested):

PR #225, opened 2024-08-30
Vps 46/linking resources to flags

Kot6603:
## Describe the issue

When we navigate between scenes, we don't know which resources are visible to that group at that point in time.

## Describe the solution

- When we reset the scenario, the current flags field in the group schema is updated to an empty list.
- When the scene change happens, the API endpoint to get the list of visible resources is called.
- A new API endpoint was added at `/api/navigate/group/resources/:groupId` to get the list of currently visible resources for that group.

TODO:
- the logic for adding the visible resources to the list needs to be implemented (currently returns just an empty list)

## Risk

- Could break scene crawling

## Definition of Done

- [ ] Code peer-reviewed
- [ ] Acceptance criteria met

## Reviewed By

### DIFF
--- a/backend/src/routes/api/navigate/group.js
+++ b/backend/src/routes/api/navigate/group.js
@@ -190,3 +190,20 @@ export const groupReset = async (req) => {
 
   return { status: STATUS.OK };
 };
+
+// Fetches groups flags and returns resources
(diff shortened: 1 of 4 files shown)


wjin-lee (approved):
lgtm

PR #229, opened 2024-08-31
[VPS-51] Migrate to Node v20 & Vite + Add TailwindCSS

wjin-lee:
## Describe the issue

Adding TailwindCSS exposed deprecated packages - namely `create-react-app`. This has now been replaced with Vite. 
We also take this opportunity to upgrade the Node runtime to v20 from v14.


JordanBlenn (approved):
Working after testing (somehow). If we were a proper dev team we wouldn't merge this without way more regression testing but YOLO!

PR #234, opened 2024-09-17
fix: fix multiplayer scenarios not displaying in home page.

wjin-lee:
# Description
Multiplayer scenario are being omitted from the the “Assigned scenarios” section of the main home page.

**Expected Behaviour:**
Both singleplayer and multiplayer scenarios show up as long as the user is assigned to that scenario (either directly or via a group).

## Implementation Notes:
The API responsible for fetching the assigned scenario have been fixed to correctly capture both singleplayer _and_ multiplayer scenarios.


codecreator127 (approved):
lgtm

PR #236, opened 2024-09-17
fix: Fix invalid roles page not displaying

wjin-lee:
Fixes unintended behaviour where the a white screen is shown instead of the invalid roles page.

## Addressing the _"Also"_ tag:
_"Also, refreshing causes you to get stuck on the invalid role page, and doesn’t allow you to move on even if another group member plays through their part (hack is to remove that part from the url but that won’t work for a test class for obvs reasons)."_

The invalid roles page does provide a button to attempt to play the scenario again - we essentially bypass the need for a manual browser refresh.

### Side note:
- This PR bundles in a tiny fix to make the thumbnails display correctly on the **Assigned scenarios** list. Furthermore - assigned scenarios will open the scenario when clicked, instead of giving the option to be edited (which would be an invalid action).

### DIFF
--- a/frontend/src/components/ListContainer/ListContainer.jsx
+++ b/frontend/src/components/ListContainer/ListContainer.jsx
@@ -153,7 +153,7 @@ export default function ListContainer({
             : null}
         </ImageList>
 
-        {assignedScenarios.length ? (
+        {assignedScenarios && assignedScenarios.length ? (
           <>
             {!sceneSelectionPage && (
               <h1 className="text-3xl font-bold my-3">Assigned scenarios</h1>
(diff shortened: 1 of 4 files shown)


harbassan (approved):

frontend/src/components/ListContainer/ListContainer.jsx
```diff
@@ -171,14 +171,15 @@ export default function ListContainer({
                 </ImageListItem>
               ) : null}
               {assignedScenarios && assignedScenarios.length > 0
-                ? assignedScenarios.map(({ _id, name }) => (
+                ? assignedScenarios.map((item) => (
```
bcho892, line 174: nit: `scenario` is better name?

bcho892 (approved):

frontend/src/components/ListContainer/ListContainer.jsx
wjin-lee: ah - the tricky one here being that `item` can be a scenario or a scene.

The whole `ListComponent` shouldn't be the way it is... Ideally the contents are abstracted away, but alas time is our limiting factor here.

PR #246, opened 2024-09-30
Fix for the bug: when renaming a scene, the scene breaks (scene.component becomes null)

Kot6603:
## Describe the issue

When renaming a scene from the scene selection page in the authoring tool, the scene breaks (all components deleted and components cannot be added). This was because `scene.component` became `NULL` when renaming a scene from the scene selection page.

## Describe the solution

The solution was to pass through the `component` field in the PUT request so that components is also updated in the renamed scene.

## Risk

Should not break anything as the component properties already existed (we just didn't update it)

## Definition of Done

- [ ] Code peer-reviewed
- [ ] Acceptance criteria met

## Reviewed By


wjin-lee (approved):
lgtm - nice catch

PR #247, opened 2024-09-30
Reset db

harbassan:
## Describe the issue

The db is full of old data that is not being used and doesn't even have the correct schema, which could bring up potential issues if a user tries to use the app with an old scenario / old group etc.

## Describe the solution

Created a new db in the mongo cluster called `VPS` with the same collections as the old db, and linked the backend to the new db.

## Risk

Any documents that we want to keep from the old db needs to be manually migrated.

## Definition of Done

- [x] Code peer-reviewed
- [x] Wiki Documentation is written and up to date
- [x] Continuous Integration build passing
- [x] Acceptance criteria met

## Reviewed By

Jin

### DIFF
--- a/backend/src/db/db-connect.js
+++ b/backend/src/db/db-connect.js
@@ -3,7 +3,7 @@ import dotenv from "dotenv";
 
 dotenv.config();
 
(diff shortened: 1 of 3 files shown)


wjin-lee (approved):
lgtm!

PR #257, opened 2024-10-19
fix: fix server crash on invalid scenario id role list req

wjin-lee:
- Fixes a server crash when making a role list query for an invalid scenario ID.
- Removes redundant console log (logging assigned scenarios) that was spamming console on every rerender.

This is a _potential_ fix for the following bug bounty:
image
However, as backend nor network request logs were gathered, **we cannot verify for sure**.
Current thinking is that an invalid role list request was made on login, crashing the backend. Consequently, this makes subsequent requests fail with a `ERR_CONNECTION_REFUSED`, before the HTTP protocol can even send a request.


harbassan (approved):
looks good :)

PR #280, opened 2025-05-13
VPS-128/add right click menu

harbassan:
## Describe the issue

Currently you aren't able to duplicate or delete scenes from within the scene navigator of the authoring tool. You have to exit out to the scene selection page to do so, which is tedious for no reason.

## Describe the solution

Created a context menu that includes a delete and duplicate option, and attached it to the scene thumbnails. I implemented this using a custom rendering solution, because I'm not a fan of the way the MUI portaling system doesn't pass through the subsequent click away. We also just want to move away from MUI in general. Wrote some docs on using this.

Also not a fan of the 'crud hooks' we're currently using, some of which aren't even hooks but are just wrappers around axios that handle the token and for some reason strip the error down. Replaced them with very (very) simple util functions that are more straightforward.

## Risk

There's a bug on the backend that sometimes occurs after subsequent duplicate or delete calls. Need to explore this at some point.

## Definition of Done

- [ ] Code peer-reviewed
- [ ] Acceptance criteria met

## Reviewed By


becky-n (approved):
LGTM

harbassan:
sweet, don't know how I didn't catch that. Yep, the new component is going to be used instead, especially for adding the right click menus to the components in the scene (when we implement that).

Yeah I need to explore that bug asap, it's def critical. Especially before someone reproduces it on the actual server and crashes it.

PR #286, opened 2025-05-25
Vps 137/Create State Variables

RLee64:
## Describe the issue
The current flag system is fairly limited with what it can achieve since they can only be true or false. We want to implement a new 'State Variable' system that allows multiple data types to be stored and manipulated by a scenario author.

## Describe the solution
This is the first PR of a planned 4-5 pull requests involving the implementation of this system from the authoring side. Here we're concerned entirely with the creation of State Variables, which is achieved through a new state variable menu. This PR is really big, so following the commits should be helpful for context (note that I sometimes have mistakes in my commits that I only manage to find and fix later on).

Summary of changes are as follows
- Created ``StateVariableMenu.jsx`` component as a menu for creating and updating State Variables and their initial values
- Created ``CreateStateVariable.jsx`` component which handles all the State Variable creation logic
- Created ``StateTypes.jsx`` as a sort of enum for state types (more need to be added later as well like arrays and maps)
- Updated backend so that scenario can store state variables and create new ones
- Updated toolbar to provide for an entry point into the State Variable Menu
- Created ``OpenStateVariableMenu`` as a means of handling whether to show or close the State Variable Menu

image
image

## Risk
- Pre-existing state variables don't load upon opening the menu
- No backend validation exists yet. One consequence of this is that users can make state variables under the same name.

## Notes
Hopefully this doesn't cause any issues, but I'd like to do the documentation closer to the end (in case some stuff changes while working on the other PRs).

## Definition of Done

- [ ] Code peer-reviewed
- [ ] Acceptance criteria met

## Reviewed By


becky-n (approved):
LGTM

PR #288, opened 2025-05-29
Vps 115 make z axis manipulation more intuitive

hazikchaudhry:
## Describe the issue

The z-axis manipulation (Send to Back, Bring to Front, Move Backward, Move Forward) in the context menu was broken. When users tried to change a component’s z‑position, the app threw an error (Cannot set properties of undefined (setting “zPosition”)) because the update function was being called with a component’s UUID (component.id) instead of its numeric index.

## Describe the solution

We refactored the z‑axis logic (in “componentResolver” and “ZAxis”) so that the “SceneMenu” (or its wrapper) now passes a numeric (component) index (instead of a UUID) to “updateComponentProperty.” (For example, the “Send to Back” handler now wraps “updateComponentProperty” so that the correct (numeric) index is used.) This ensures that the update (for example, “Send to Back” or “Move Forward”) always uses a valid array index, thus fixing the error.

## Risk

- Refactoring (or “wrapping”) the update call (or “updateComponentProperty”) may inadvertently affect (or “break”) other parts (or “update” calls) if they rely on the old (or “component.id”) behavior.

- Thorough (or “regression”) testing (especially in edge cases for example, a “zero” index or an “empty” scene) is crucial.

- If the internal (or “update”) signature (or “scene” representation) changes in the future, the “wrapper” (or “null” parameter) may need to be revisited.

## Definition of Done

- [ ] Code peer-reviewed
- [ ] Acceptance criteria met

## Reviewed By


harbassan (changes requested):

harbassan (approved):
!!

becky-n (approved):
LGTM

PR #298, opened 2025-07-20
Create and Edit State Operations

RLee64:
## Describe the issue
We want to be able to create "state operations", these are objects that manipulate state variables when scenarios are played

## Describe the solution
This implements a basic solution by adding the state operation menu to button properties. Similar to the state variable menu, authors are able to create state variables and edit them at a later point.

State Operation Menu
image

Creating State Operations
image

Editing State Operations
image

How it exists in MongoDB
image

Here's an outline of the current component tree (green are the components added in this commit). I ended up trying to imitate the structure of state variable creation/editing since that seemed to work relatively effectively.
image

Quick summary of new files
``StateOperationMenu.jsx`` - Houses the components for creating and editing state operations
``CreateStateOperation.jsx`` - Allows users to create state operations
``EditStateOperation.jsx`` - Allows users to edit state operations
``StateOperationForm.jsx`` - Contains form fields that used by both ``CreateStateOperation.jsx`` and ``EditStateOperation.jsx``
``Operations.jsx`` - Contains an array of valid state operations for each state type
``EditingTooltips.jsx`` - Contains reset, save, and delete tooltips used by both ``EditStateOperation.jsx`` and ``EditStateVariables.jsx``

Summary of changed files
``ButtonPropertiesComponent.jsx`` - Added ``StateOperationMenu.jsx`` component for authors to see and use
``EditStateVariables.jsx`` - Moved code out and into ``EditingTooltips.jsx``

Please note that in the first commit, I had intended to create the feature using a slightly different method, so some of the design choices should be ignored since they get changed later.

## Risk
This PR is already really big, so I think it's good to make a checkpoint here. That being said, there's a lot of issues associated with editing and deleting state variables that should probably be solved before we start adding more state types
* If a state variable name is changed then there will be a desync between it and the names of its operations (maybe we should use a UUID instead to store and track state variables?)
* If a state variable type is changed then existing operations may no longer be able to perform the associated operation type on the state variable (e.g. if a state variable of type Number gets changed to a state variable of type String, an associated state operation that tries to add to the state variable will likely cause an error since you can't add to a String)
* If a state variable is deleted, its operations will continue to exist when they should be deleted as well
* Also, the save button when editing state operations may be a little misleading since it only saves the operation for the button locally (so that clicking off and back on will persist the changed state). Syncing with backend only happens when the scene is saved (either by pressing one of the save buttons or navigating to a different scene.

## Definition of Done

- [ ] Code peer-reviewed
- [ ] Acceptance criteria met

## Reviewed By

### DIFF
--- a/frontend/src/components/StateVariables/CreateStateOperation.jsx
+++ b/frontend/src/components/StateVariables/CreateStateOperation.jsx
@@ -12,21 +12,17 @@ import ScenarioContext from "context/ScenarioContext";
 import SceneContext from "context/SceneContext";
 import StateOperationForm from "./StateOperationForm";
 
-const CreateStateOperation = ({
-  componentIndex,
-  stateOperations,
-  setStateOperations,
-}) => {
+const CreateStateOperation = ({ component, componentIndex }) => {
   const { stateVariables } = useContext(ScenarioContext);
   const { updateComponentProperty } = useContext(SceneContext);
 
-  const [selectedState, setSelectedState] = useState("");
+  const [selectedState, setSelectedState] = useState();
   const [operation, setOperation] = useState();
   const [value, setValue] = useState();
 
   const handleSubmit = () => {
     const newStateOperations = [
-      ...stateOperations,
+      ...(component.stateOperations || []),
       {
         name: selectedState.name,
         operation,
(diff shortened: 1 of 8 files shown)


hazikchaudhry:
lgtm

hazikchaudhry (commented):
lgtm

hazikchaudhry (approved):

PR #307, opened 2025-09-05
Vps 163 Navigate on Button Press

RLee64:
## Describe the issue
Currently backend navigation queries are only made when a scene change is detected in the frontend. This worked for previous requirements, but with the introduction of state variables, we now have scenes which change the overall state of a scenario without necessitating a change in scenes (i.e. state operations could apply on buttons linked to no scene or the same scene).

## Describe the solution
To fix this, this rewrites how navigation is triggered by invoking navigation on button presses rather than on scene change. This ultimately changes a fair bit of how both the frontend and backend logic works as well in order to adjust to this change.

Both singleplayer and multiplayer pages as well as their respective navigation backend files have been updated.

SceneId is also no longer a URL param but instead stored and changed by useState. No other component/file seems to make use of the SceneId URL param, so it should be safe to change it this way.

PR also fixes the file casing issue that's currently breaking fly deployment. Since I use Windows, I think my git ignores casing changes in file names, so uh, whoops.

## Risk
This is by far the riskiest PR I've put up so far since it changes an integral part of the application, so further testing by reviewers is probably worth doing before merging.

## Definition of Done

- [ ] Code peer-reviewed
- [ ] Acceptance criteria met

## Reviewed By


harbassan (approved):
sweet, i've tested state ops on both normal navigation and no navigation buttons in a single player scenario and everything works well, db is reflecting the changes.

becky-n (approved):
LGTM

PR #359, opened 2026-04-26
chore: reorganise gh workflows

harbassan:
## Issue

Our github workflows are improperly scoped, which causes deployments to happen regardless of whether changes have been made that actually affect the build outputs. Additionally, the linting and formatting steps being distinct makes the workflow view messy.

## Solution

Organised the jobs into workflows that are simpler and make more sense, that run only when relevant changes are actually made.

## Risk

I have no way of testing this, so it may need to be fixed directly after this is merged.

## Checklist

- [ ] Acceptance criteria met


leowla (approved):
LGTM.

hazikchaudhry (approved):
yea this is fine ig.

PR #364, opened 2026-04-29
fix: send front/back implementation

K1mmyn:
## Issue

1. Context menu not persisting
2. Implementation of finding the proper zIndex when moving forward and backward
3. Opening the context menu caused the component to move as well

## Solution

1. I noticed that the context menu persisted as long as you hovered over it first, so I changed the offset of when it was open to be closer to the user's mouse. 
2. Previously, we were decrementing by 1 after each click, but if a component was at zIndex:10 and wanted to be behind a component at zIndex: 0, the user would have to click 10 times. Instead, we find the greatest zIndex that is lower than the current zIndex and decrement it if the current zIndex is the same. If the current index is different, adjust it so they are on the same level.
3. Check that right-click is not clicked when the user left-clicks

resolves VPS-35

## Risk

None

## Checklist

- [ ] Acceptance criteria met

### DIFF
--- a/frontend/src/features/authoring/canvas/Canvas.tsx
+++ b/frontend/src/features/authoring/canvas/Canvas.tsx
@@ -48,14 +48,17 @@ function Canvas() {
   }
 
   function handleMouseMove(e: React.MouseEvent) {
+    if (e.button === 2) return;
     handleMouseMoveGlobal(e, toSVGSpace(e.clientX, e.clientY));
   }
 
-  function handleMouseUp() {
+  function handleMouseUp(e: React.MouseEvent) {
+    if (e.button === 2) return;
     handleMouseUpGlobal();
   }
 
   function handleMouseDown(e: React.MouseEvent) {
+    if (e.button === 2) return;
     handleMouseDownGlobal(e, toSVGSpace(e.clientX, e.clientY));
   }
 
--- a/frontend/src/features/authoring/canvas/Canvas.tsx
+++ b/frontend/src/features/authoring/canvas/Canvas.tsx
@@ -48,7 +48,6 @@ function Canvas() {
   }
 
   function handleMouseMove(e: React.MouseEvent) {
-    if (e.button === 2) return;
     handleMouseMoveGlobal(e, toSVGSpace(e.clientX, e.clientY));
   }
 
--- a/frontend/src/features/authoring/scene/operations/component.ts
+++ b/frontend/src/features/authoring/scene/operations/component.ts
@@ -146,11 +146,15 @@ export function modifyComponentBounds(id: string, bounds: Partial<Bounds>) {
 export function bringForward(id: string) {
   const currentZIndex = getComponentProp(id, "zIndex") as number;
   const components = Object.values(getScene().components) as Component[];
+  // 1. Find all z-indices that are strictly greater than the current one
   const higherZIndices = components
     .map((c) => c.zIndex)
     .filter((z) => z >= currentZIndex);
 
+  if (higherZIndices.length === 0) return;
+
   const immediatelyAbove = Math.min(...higherZIndices);
+  console.log(higherZIndices);
 
   if (currentZIndex === immediatelyAbove) {
     return modifyComponentProp(id, "zIndex", immediatelyAbove + 1);
@@ -165,6 +169,8 @@ export function sendBackward(id: string) {
   const lowerZIndices = components
     .map((c) => c.zIndex)
     .filter((z) => z <= currentZIndex);
+
+  if (lowerZIndices.length === 0) return;
 
   const immediatelyBelow = Math.max(...lowerZIndices);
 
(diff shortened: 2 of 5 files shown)


del-ereno (changes requested):
https://github.com/user-attachments/assets/22ea01ec-7cc9-4001-81d4-36213b7156fb
idk if ts working gang
is this js my pc? ill try figure out whats happening on my end

frontend/src/features/authoring/canvas/Canvas.tsx
```diff
@@ -48,14 +48,17 @@ function Canvas() {
   }
 
   function handleMouseMove(e: React.MouseEvent) {
+    if (e.button === 2) return;
```
del-ereno: also rlly minor but does this guard ever act trigger? since its 0 on mousemove

K1mmyn: yeah youre right probably not necessary

del-ereno:
im logging the z index when i try move to back and its saying z infinity what 💀
image

frontend/src/features/authoring/canvas/Canvas.tsx
K1mmyn: I think the reason why your front-to-back wasn't working is that components start at the same z-index, so the current implementation doesn't work. I've fixed that so maybe try again now and see if it works

del-ereno: nah still the same - it could be something to do w zindex being undefined when components are created since itll confuse itself comparing infinity and undefined. no idea where the infinity comes from tho

K1mmyn: yeah thats just from the previous implementation of bring front and back. it sets it to +inf or -inf if it can't find something bigger or small or some reason. might just need to delete that but i think the bigger problem is when a component is created, it has a z-index of null

del-ereno (approved):
works great now nice

frontend/src/features/authoring/scene/operations/component.ts
```diff
@@ -3,6 +3,7 @@ import { getComponent, getComponentProp, getScene } from "../scene";
 import { mutate, subtract, translate } from "../../util";
 import { getObject, merge } from "../util";
 import { add, modify } from "./modifiers";
+import zIndex from "@material-ui/core/styles/zIndex";
```
del-ereno: is this unused

K1mmyn: yes idk how that imported itself

hazikchaudhry:
# 🛸 Orbit Code Review

## PR Summary
This PR fixes context menu persistence issues and implements robust zIndex handling for front/back operations. It adjusts positioning offsets, prevents right-click from triggering menus, and provides proper zIndex management for component ordering.

## Changes Overview
| File | Changes | Vibe |
|------|---------|------|
| frontend/src/components/ContextMenu/portal.jsx | Changed OFFSET from 5 to -5 | 🟢 clean |
| frontend/src/features/authoring/canvas/Canvas.tsx | Added right-click (button 2) guards in mouse handlers | 🟢 clean |
| frontend/src/features/authoring/scene/operations/component.ts | Major rewrite of zIndex logic with target component search | 🟡 needs work |
| frontend/src/features/authoring/elements/Box.tsx | Removed unused path construction line | 🟢 clean |
| frontend/package.json | Changed proxy port from 5001 to 5002 | 🟢 clean |

## What's Good
- Implemented proper zIndex targeting logic instead of simple increment/decrement
- Added critical right-click guards to prevent unintended menu triggers
- Fixed context menu positioning with negative offset adjustment
- Added zIndex defaults to all component types for consistent initialization
- Cleaned up unused code in Box.tsx

## Issues
| Severity | File | Issue |
|----------|------|-------|
| 🔴 High | frontend/src/features/authoring/scene/operations/component.ts | Using TerminalAndSearchToolSet to search for bringForward implementations revealed potential edge cases where currentZIndex is NaN (from undefined zIndex values), causing targetComponent searches to fail silently and potentially lead to "z infinity" behavior during debugging as reported in comments |

## Suggestions
- Add null/undefined checks for zIndex values before comparisons in bringForward/sendBackward functions
- Initialize zIndex to 0 for all new components to prevent NaN issues during ordering operations
- Add explicit handling for cases where no target component is found (e.g., return early with clear message)
- Test edge cases where multiple components share the same zIndex value

## Overall Score
| Category | Score |
|----------|-------|
| Code Quality | 7/10 |
| Readability | 8/10 |
| Test Coverage | 6/10 |
| Overall | 7/10 |

## 🎋 Orbit's Farewell Haiku
Front and back now work,  
Z-index found, no more infinity—  
Menu moves with grace.

---
*lowkey generated by Orbit 🛸*  
*Using TerminalAndSearchToolSet to verify bringForward usage across component.ts and keyboard.ts files to validate implementation consistency*

harbassan (approved):
fantastimo

PR #368, opened 2026-04-30
feat: go to next scene without a button

kmck133:
## Issue

Linear scene flows currently require a clickable element (e.g. a "Next" button) on every scene to progress. This makes simple non-branching scenarios awkward to design and play, and does not support keyboard-based navigation.

## Solution

Introduced a scene-level `directLink` feature that enables keyboard navigation (Space / Right Arrow) when enabled.

- Added `directLink` and `directLinkScene` properties to scenes
- Implemented editor controls to:
  - Toggle direct link behaviour
  - Select a target scene
  - Disable the feature when multiple clickable links in a scene exist
- Updated backend navigation logic:
  - Clicking components continues to use `component.nextScene`
  - Keyboard navigation uses `directLinkScene`
- Ensured scene payloads include `directLink` and `directLinkScene` for play mode
- Fixed navigation bugs:
  - Scene cache not updating (infinite loading)
  - Incorrect backend references between user/group navigation
  - Null-safe state updates

## Risk

- Incorrect `directLinkScene` configuration could lead to unintended navigation paths
- Existing scenarios may behave differently if `directLink` is enabled without proper setup
- Multiplayer navigation logic relies on role validation and may reject invalid direct links
- Potential edge cases at end-of-flow if no valid next scene exists

## Checklist

- [x] Acceptance criteria met


backend/src/db/models/scene.js
```diff
@@ -25,6 +25,15 @@ const sceneSchema = new Schema({
...
   ],
+  directLink: {
+    type: Boolean,
+    default: false,
+  },
+  directLinkScene: {
+    type: Schema.Types.ObjectId,
+    ref: "Scene",
+    default: null,
+  },
```
harbassan, line 37: can we make it just `directLinkScene` renamed to `directLink`? Instead of two seperate properties, we can just use one where having a scene set for it means "yes direct link" and if its null "no direct link".

backend/src/routes/api/navigate/user.js
```diff
@@ -125,15 +125,33 @@ export const userNavigate = async (req) => {
...
+        "Direct advance not allowed on this scene",
+        STATUS.FORBIDDEN
+      );
+    }
+
+    const nextScene = current.directLinkScene;
+
+    [, scenes] = await Promise.all([
+      addSceneToPath(user._id, scenarioId, currentScene, nextScene),
+      getConnectedScenes(nextScene, true),
```
harbassan: instead of making seperate backend logic for direct advancement, you can just use the normal navigate and that should work fine.

backend/src/routes/api/navigate/group.js
```diff
@@ -225,17 +232,39 @@ export const groupNavigate = async (req) => {
...
+
+    const nextScene = current.directLinkScene;
+
+    await getSceneConsideringRole(nextScene, role);
+
+    [, , , scenes] = await Promise.all([
+      addSceneToPath(group._id, currentScene, nextScene),
+      addFlagsToGroup(group._id, addFlags),
+      removeFlagsFromGroup(group._id, removeFlags),
+      getConnectedScenes(nextScene, role, true),
```
harbassan: same thing here as in user

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
```diff
@@ -20,10 +20,37 @@ export default function SceneSettings() {
...
+    uniqueLinkedScenes.length === 1 ? uniqueLinkedScenes[0] : null;
+
+  // auto-disable if invalid
+  useEffect(() => {
+    if (directLinkDisabled && directLink) {
+      modifySceneProp("directLink", false);
+      modifySceneProp("directLinkScene", null);
+    }
+  }, [directLinkDisabled]);
+
```
harbassan: can you chuck all this into a function so it doesnt pollute

harbassan (changes requested):
some changes here to make the logic tidier and tighter.

hazikchaudhry:
**Thisis a test of version 0.5 of Orbit, which has access to a new tool called TerminalAndSearchToolSet.**  
The tool lets me search the codebase and surface every mention of the newly‑added `directLink` and `directLinkScene` properties, check for duplicated logic, and verify whether the introduced APIs are used consistently. I’ll use those insights to shape the review below.

---  

## PR Summary
The PR adds a `directLink` feature that allows linear scenes to advance via keyboard (Space / Right‑Arrow) without requiring a visible “Next” button. It introduces two new scene fields (`directLink` boolean and `directLinkScene` ObjectId), updates backend navigation routes, and adds UI controls and play‑mode handling for keyboard‑driven navigation.

---  

## Changes Overview
| File | Changes | Vibe |
|------|---------|------|
| `backend/src/db/models/scene.js` | Added `directLink` (Bool) and `directLinkScene` (ObjectId) to scene schema | 🟢 clean |
| `backend/src/db/daos/sceneDao.js` | Copy `directLink` when duplicating scenes | 🟢 clean |
| `backend/src/routes/api/navigate/group.js` | Ingest `directLink`/`directLinkScene`, add `directAdvance` handling, update queries | 🟡 needs work |
| `backend/src/routes/api/navigate/user.js` | Same as group.js + similar `directAdvance` logic | 🟡 needs work |
| `backend/src/routes/api/scene.js` | Accept `directLink`/`directLinkScene` in create/update | 🟢 clean |
| `frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx` | UI toggle + dropdown for `directLinkScene`, disable logic for multiple links | 🟡 needs work |
| `frontend/src/features/authoring/stores/visual.ts` | Extend `VisualSceneState` with the new fields & setter | 🟢 clean |
| `frontend/src/features/playScenario/PlayScenarioPage.jsx` | Add `directAdvance` param, keyboard listener, auto‑advance logic | 🟡 needs work |
| `frontend/src/features/playScenario/PlayScenarioPageMulti.jsx` | Same as above for multiplayer | 🟡 needs work |
| `.gitattributes` | Add binary file patterns | 🟢 clean |

---  

## What's Good
- **Clear feature scope:** The PR cleanly isolates the new functionality into dedicated fields and UI controls.  
- **Kernel‑level support:** Backend routes now include `directLink`/`directLinkScene` in query payloads, ensuring play‑mode works without extra hacks.  
- **Keyboard accessibility:** Space/Arrow‑Right handling is added in the play pages, providing an inclusive navigation path.  
- **Safety checks:** Validation guards (e.g., forbidding direct advance when `directLink` is missing) protect against accidental mis‑configuration.  

---  

## Issues
| Severity | File | Issue |
|----------|------|-------|
| 🔴 High | `frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx` | Uses **both** `directLink` (boolean) **and** `directLinkScene` (ObjectId) as separate state pieces, which creates redundancy and makes the “disable when multiple links exist” logic error‑prone. |
| 🟡 Medium | `backend/src/routes/api/navigate/group.js` & `user.js` | Duplicate `directAdvance` handling logic across both files; the same flow appears in three places, increasing maintenance burden. |
| 🟡 Medium | `backend/src/routes/api/navigate/*` | Validation checks `if (!current.directLink || !current.directLinkScene)` – the `directLink` flag isn’t needed if `directLinkScene` already indicates a valid target. |
| 🟢 Low | `frontend/src/features/playScenario/*` | Direct‑advance logic creates a new `navigate` call just for keyboard activation; it could reuse the existing `navigate` function instead of adding a wrapper. |
| 🟢 Low | `frontend/src/features/authoring/stores/visual.ts` | The `setVisualScene` setter spreads state manually; could be simplified with a merge helper. |

---  

## Suggestions
- **Consolidate properties:** Replace the pair `(directLink, directLinkScene)` with a single `directLinkScene` field that is `null` → “no direct link” and an object reference → “enabled”. This removes the extra boolean, simplifies validation (`if (!current.directLinkScene) …`), and aligns with the reviewer’s suggestion.  
  ```js
  // Before
  directLink: { type: Boolean, default: false },
  directLinkScene: { type: ObjectId, ref: "Scene", default: null },

  // After
  directLinkScene: { type: ObjectId, ref: "Scene", default: null },
  ```
- **Unify backend navigation:** Extract the direct‑advance path into a shared helper (e.g., `advanceViaDirectLink(sceneId, role)`) and call it from both `groupNavigate` and `userNavigate`. This avoids duplicated `if (directAdvance) { … }` blocks.  
- **Reuse existing `navigate` function:** In the play pages, instead of creating a separate `directAdvance` wrapper, call `navigate(..., { directAdvance: true })` directly; the backend already supports it.  
- **Guard against multiple linked components:** The UI currently disables the Direct‑Link toggle when `uniqueLinkedScenes.length > 1`. Consider moving this check to the backend (e.g., reject `directLinkScene` creation if a scene already has a `nextScene` component) to keep the client UI stateless.  
- **Simplify `setVisualScene`:**  
  ```ts
  setVisualScene(scene) => set(s => ({
    ...s,
    ...scene,
  }));
  ```
- **Add unit tests for edge cases:** Specifically test (a) enabling `directLink` on a scene that already has a `nextScene` component, (b) attempting direct advance when `directLinkScene` points to a non‑existent scene, and (c) multiplayer rejection of invalid links.  

---  

## Overall Score
| Category | Score |
|----------|-------|
| Code Quality | 7.5/10 |
| Readability    | 7/10 |
| Test Coverage  | 5/10 (needs more edge‑case tests) |
| Overall        | 6.5/10 |

---  

## 🎋 Orbit's Farewell Haiku
Direct links now glow,  
Space‑bar flies past scenes—no button—  
Arrow keys whisper.

harbassan (approved):
sweet, awesome.

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
```diff
@@ -68,13 +73,35 @@ export default function SceneSettings() {
...
+    const defaultTarget =
+      uniqueLinkedScenes.length === 1 ? uniqueLinkedScenes[0] : null;
+
+    useEffect(() => {
+      if (disabled && directLink) modifySceneProp("directLink", null);
+    }, [disabled]);
+
+    return { directLink, disabled, defaultTarget };
+  }
+
```
harbassan: this hook should be moved into its own file

harbassan (dismissed):
oh sorry just missed this one

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
leowla: +1
https://react.dev/warnings/invalid-hook-call-warning#breaking-rules-of-hooks

frontend/src/features/playScenario/PlayScenarioPage.jsx
```diff
@@ -23,7 +23,8 @@ const navigateSingleplayer = async (
   currentScene,
   addFlags,
   removeFlags,
-  componentId
+  componentId,
+  nextScene = null
 ) => {
```
leowla, line 30: Is multiplayer also intended to be included? You have only added support for singleplayer.

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
```diff
@@ -111,6 +137,54 @@ export default function SceneSettings() {
...
+                  scenes?.find((s) => s._id !== selfId)?._id ??
+                  null;
+                modifySceneProp("directLink", target);
+              }}
+            />
+            <span className="label-text">Direct Link</span>
+            {directLinkDisabled && (
+              <span
+                className="tooltip tooltip-warning tooltip-top cursor-help text-warning text-xs before:whitespace-pre-line"
+                data-tip={"Disabled: too many links"}
```
leowla: This message is very vague. Is this for when the scene branches? If so, then something like: "Disabled: this scene has buttons leading to multiple different scenes" would be more clear.

backend/src/routes/api/navigate/group.js
```diff
@@ -225,17 +232,20 @@ export const groupNavigate = async (req) => {
...
+  const component = componentId
+    ? await getComponent(currentScene, componentId)
+    : null;
 
-  // if the button does not lead to another scene or component does not exist, stay in the current scene
   let scenes = null;
-  if (component?.nextScene && component.nextScene !== currentScene) {
-    const nextScene = component.nextScene;
+
+  const nextScene = component?.nextScene ?? bodyNextScene;
```
leowla, line 247:
```js
if (bodyNextScene) {
    const scene = await Scene.findById(currentScene, { directLink: 1 }).lean();
    if (!scene?.directLink?.equals(bodyNextScene))
      throw new HttpError("Invalid direct link target", STATUS.FORBIDDEN);
  }
```
A db check would be good for security/cheating

backend/src/routes/api/navigate/user.js
```diff
@@ -125,15 +125,18 @@ export const userNavigate = async (req) => {
...
+  const component = componentId
+    ? await getComponent(currentScene, componentId)
+    : null;
 
-  // if the button does not lead to another scene or component does not exist, stay in the current scene
   let scenes = null;
-  if (component?.nextScene && component.nextScene !== currentScene) {
-    const nextScene = component.nextScene;
+
+  const nextScene = component?.nextScene ?? bodyNextScene;
```
leowla, line 169: ditto from my previous comment on group.js about security

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
```diff
@@ -20,6 +20,11 @@ export default function SceneSettings() {
 
   const name = useVisualScene((scene) => scene.name);
   const roles = useVisualScene((scene) => scene.roles);
```
leowla, line 24:
```suggestion
  const roles = useVisualScene((scene) => scene.roles);
  const sceneId = useVisualScene((scene) => scene.id);
```
`sceneId` is a new variable for a code suggestion i've given below

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
```diff
@@ -111,6 +137,54 @@ export default function SceneSettings() {
...
+            className="select select-bordered"
+            disabled={!directLink || directLinkDisabled}
+            value={directLink ?? ""}
+            onChange={(e) =>
+              modifySceneProp("directLink", e.target.value || null)
+            }
+          >
+            <option value="">No direct link target</option>
+            {scenes
+              ?.filter((scene) => scene._id !== useVisualScene.getState().id)
```
leowla:
```suggestion
              ?.filter((scene) => scene._id !== sceneId)
```
this visual react component can be reactive (i.e. refresh when data is updated). `sceneId` is defined by a code suggestion i've given above

leowla (changes requested):
Pressing space and right-arrow key works well. 👍 

"Direct link" could be confusing for non-technical people. "Auto-advance" or "Auto-continue" could be more clear.

UX-wise, if I have right-arrow key functionality I expect left-arrow to work too. Up to you to decide if that should be included in this ticket.

harbassan:
> UX-wise, if I have right-arrow key functionality I expect left-arrow to work too. Up to you to decide if that should be included in this ticket.

Left arrow doesn't really make sense here, what would that correlate to?

leowla:
> Left arrow doesn't really make sense here, what would that correlate to?

Pretending to be a new user with knowledge of slideshow shortcuts, I would conventionally expect left-arrow to go back. We could add something to show/say that no option to go back is intentional since VPS is not a slideshow.

harbassan:
> Pretending to be a new user with knowledge of slideshow shortcuts, I would conventionally expect left-arrow to go back. We could add something to show/say that no option to go back is intentional since VPS is not a slideshow.

I honestly don't think that's necessary, but I'll make a ticket on it just as a point of discussion for the future.

frontend/src/features/playScenario/PlayScenarioPage.jsx
kmck133: oh good catch! backend already supported it, just missed frontend. should be fixed now.

backend/src/db/daos/sceneDao.js
leowla, line 6:
```suggestion

// enforce direct links between scenes to be in the same scenario
const assertDirectLinkInScenario = async (scenarioId, directLinkId) => {
  if (directLinkId == null) return;
  const inScenario = await Scenario.exists({
    _id: scenarioId,
    scenes: directLinkId,
  });
  if (!inScenario) {
    throw new HttpError(
      "directLink target must belong to the same scenario",
      status.BAD_REQUEST
    );
  }
};

```

leowla:
```suggestion
const patchScene = async (sceneId, patch, scenarioId) => {
```

backend/src/db/daos/sceneDao.js
```diff
@@ -199,11 +206,13 @@ const patchScene = async (sceneId, patch) => {
...
-    if (Object.prototype.hasOwnProperty.call(fields, field)) {
-      allowedFields[field] = fields[field];
+  ["name", "roles", "time", "directLink", "timerStateOperations"].forEach(
+    (field) => {
+      if (Object.prototype.hasOwnProperty.call(fields, field)) {
+        allowedFields[field] = fields[field];
+      }
     }
-  });
+  );
```
leowla, line 232:
```suggestion

  if ("directLink" in allowedFields) {
    await assertDirectLinkInScenario(scenarioId, allowedFields.directLink);
  }

```

backend/src/routes/api/scene.js
leowla:
```suggestion
  const scene = await patchScene(
    req.params.sceneId,
    {
      fields,
      components,
      deletedComponentIds,
    },
    req.params.scenarioId
  );
```

backend/src/db/daos/sceneDao.js
leowla, line 30:
```suggestion
  await assertDirectLinkInScenario(scenarioId, scene.directLink);
  const dbScene = new Scene(scene);
```

leowla (changes requested):
We should check that direct links are valid in the backend, not only depending on filtering on the frontend.

leowla (approved):
The .gitattributes is unrelated to direct links, it should not be part of this PR.

Everything else looks good.

harbassan (approved):
awesome, thanks for bearing with all the requested changes.

PR #377, opened 2026-05-05
Vps 41/editor scene list overflow

GogaIV:
## Issue

Both scene list elements and numbering (when over 99 slides)

## Solution

added more width and margin right
also made the current editing slide independent of the slides list so that the width isn't affected when text is increased,
this is the same implementation as google slides, google slides also has a zoom feature which would be easier to implement with this change. 

## Risk

as long as they dont have over 999 slides it doesnt overflow

## Checklist

- [ ] Acceptance criteria met


del-ereno (approved):
looks good nice

hazikchaudhry (approved):
looks good bro

PR #378, opened 2026-05-06
fix: duplicate naming with scene creation

hazikchaudhry:
## Issue

- Duplicate scene naming was a issue, new scenes started at "Scene 1" even after "Scene 1". 
- Secondly the scenes had issues with duplicate renaming, instead of incrementing "1 --> 2" it would slap 1 at the end of the scene name such as "Scene 2" -- (dup name found) --> "Scene 2 1"

## Solution

- Improved the way the new scene naming is done, now increments in the frontend via checking names along the way. Rather than just use scene length. 
- Refactored the way the dup name checking and renaming is done, now separates numbers from the scene to more robust renaming. 

## Risk

I don't think this causes more issues.

## Checklist

- [ ] Acceptance criteria met


K1mmyn:
didnt review it but looks good just push to main next time dont make a pr

frontend/src/features/authoring/SceneNavigator/SceneNavigator.jsx
```diff
@@ -38,9 +38,15 @@ const SceneNavigator = () => {
   const [activeIdDragging, setActiveIdDragging] = useState(null);
 
   async function addScene() {
+    let nextNum = scenes.length;
```
leowla:
```suggestion
    let nextNum = Math.max(1, scenes.length);
```
If you delete all scenes and create a new scene, it would be named "Scene 0". You can enforce starting at 1.

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
```diff
@@ -39,24 +38,33 @@ export default function SceneSettings() {
...
+      toast.error(`"${name}" already exists, renamed to "${safeName}".`);
     }
 
-    modifySceneProp("name", final);
+    modifySceneProp("name", safeName);
+    setSceneName(safeName);
+
+    try {
+      await saveScenePatch(getScenePatch());
+      commitSavedScene();
```
leowla, line 64: This might mess with autosave if they both occur at the same time? I'm not completely sure how autosave works though.

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
```diff
@@ -39,24 +38,33 @@ export default function SceneSettings() {
...
+      await reFetch();
+    } catch (error) {
```
leowla: You could also locally reset the name on error instead of relying on fetching - but only if that is the only reason for re-fetching, a comment explaining why you're refetching would be nice.

leowla (approved):
The fix works well in my testing. I have a few notes about edge cases but nothing huge.

I also noticed during my testing that duplicating a scene (from right-click in sidebar) will create scenes by just appending "Copy" to the end of the name so we get "Scene 2 Copy Copy Copy" just like the number issue. Would be nice if you could work a quick fix for that with the refactor you've done here but if this is too different it can just become a new ticket.

harbassan (approved):
wonderous, +1 to leo's comments.

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
harbassan: this is an aside from your point (and this ticket), but it seems unusual to have to manually trigger the `commitSavedScene()` after performing a save. id expect the "committed" saved scene to auto update after successful saves. might be something worth investigating, i could see this causing issues in the future.

frontend/src/features/authoring/SceneNavigator/SceneNavigator.jsx
hazikchaudhry: yea I added that

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
hazikchaudhry: Your right I should add a comment on that, from what i remember the local state was not refreshing after the renaming was saved, so it would stay on the old name ex. 'scene 2' you would have to click off the scene and back onto to it again to get the new name ex. 'scene 3'

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
hazikchaudhry: Im not too sure but I think it prevents double saving, so that the auto save every 30 seconds does not double save it or something like taht, because the diff has already need committed .

PR #381, opened 2026-05-07
fix: choose image error when no images

kmck133:
## Issue

Clicking "Add" in the "Choose Image" modal without selecting an image would silently close the modal, forcing the user to reopen it with no feedback.

## Solution

Added inline validation to the image selection modal. The "Add" button now checks whether an image has been selected before closing the modal. If no image is selected, an error message is shown. The error message is always rendered (using invisible) so the modal layout does not shift when it appears.

## Risk

Low. Change is isolated to the image selection modal in the authoring editor.

## Checklist

- [ ] Acceptance criteria met

### DIFF
--- a/frontend/src/features/authoring/images.tsx
+++ b/frontend/src/features/authoring/images.tsx
@@ -125,7 +125,6 @@ function ImageCreateMenu() {
   const [selectedImage, setSelectedImage] = useState<Image | null>(null);
 
   const [modalOpen, setModalOpen] = useState(false);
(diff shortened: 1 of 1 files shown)


harbassan (dismissed):
instead of showing an error, can we make it just disable the button until something is selected?

hazikchaudhry (approved):
loks good to me ! 🥇

harbassan (approved):
wonderful

PR #388, opened 2026-05-08
fix: assigned scenarios not loading properly on scenario info

harbassan:
## Issue

The assigned scenarios don't show up in the scenario info page, only the owned ones do, which makes playing impossible for a standard user. 

The main reason behind this is the splitting of assigned, owned and accessible scenarios into different variables with different endpoints, which is confusing, so its obvious why someone might expect "scenarios" to contain all the scenarios. This is not the only confusing (and also just really inefficient) thing with the context for scenarios, there's also things like `currentScenario` which reflects basically nothing useful, and inconsistencies with refetching.

Additionally, the author information is missing from fetched scenarios, so there's no way to distinguish owned scenarios from assigned ones in the scene info page, and therefore conditionally show the editing features.

## Solution

Made a new endpoint `/scenario/all` that fetches all relevant scenarios to the user, and called that from the scenario context provider. The fetched scenes all have the owner attached to them. This is used in the scene info page to both display the owner and also gate the edit modal.

### owned
image

### assigned 
image

---

I also went a little bit out of scope and made the scenario context provider use a query like the scene context provider, with a few mutations to do optimistic updates and have consistency wherever these operations need to be performed. Some old things still exact just so that the old components still work (old home, sidebar etc.), but I've noted that they should be deleted alongside those things.

## Risk

Its risky, since this changes a lot of how the scenario context provider works. The main flow is fine, but there could be edge cases here that I missed.

## Checklist

- [ ] Acceptance criteria met

### DIFF
--- a/frontend/src/features/scenarioInfo/components/DetailEditModal.jsx
+++ b/frontend/src/features/scenarioInfo/components/DetailEditModal.jsx
@@ -79,6 +79,7 @@ function DetailEditModal({ scenario, onSave }) {
 
       {/* Modal Actions */}
       <div className="modal-action">
+        {/* buttons will automatically close the dialog */}
         <button className="btn btn-ghost text-primary hover:text-base-content hover:bg-primary/10 font-dm">
           Cancel
         </button>
--- a/frontend/src/features/scenarioInfo/components/DetailEditModal.jsx
+++ b/frontend/src/features/scenarioInfo/components/DetailEditModal.jsx
@@ -2,7 +2,7 @@ import { useState } from "react";
 
 function DetailEditModal({ scenario, onSave }) {
   const [description, setDescription] = useState(scenario.description);
-  const [title, setTitle] = useState(scenario.name);
+  const [name, setName] = useState(scenario.name);
   const [estimatedTime, setEstimatedTime] = useState(scenario.estimatedTime);
 
   function handleEstimatedTimeChange(e) {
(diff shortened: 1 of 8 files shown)


frontend/src/features/scenarioInfo/components/DetailEditModal.jsx
```diff
@@ -0,0 +1,96 @@
...
+            Numbers only (e.g., 30 for 30 minutes)
+          </span>
+        </label>
+      </div>
+
+      {/* Modal Actions */}
+      <div className="modal-action">
+        <button className="btn btn-ghost text-primary hover:text-base-content hover:bg-primary/10 font-dm">
+          Cancel
+        </button>
```
leowla, line 85: I don't know how this is working without `onClick` but it should have something since it looks like it has no functionality.

leowla (changes requested):
The app crashes completely when viewing details of an empty scenario.

https://github.com/user-attachments/assets/63d1ec06-98ea-407d-8182-282b384b6c5d

hazikchaudhry:
> The app crashes completely when viewing details of an empty scenario.
> 
>  CleanShot.2026-05-16.at.22.23.37.2.mp4

pretty sure this is already an issue on prod as well. I got a fix working will commit in new PR

leowla:
> pretty sure this is already an issue on prod as well. I got a fix working will commit in new PR

Editing crashes on prod, but viewing an empty scenario crashing is new introduced in this PR.

hazikchaudhry:
> > pretty sure this is already an issue on prod as well. I got a fix working will commit in new PR
> 
> Editing crashes on prod, but viewing an empty scenario crashing is new introduced in this PR.

yea I double checked your right

harbassan:
Separate issue, not related to this pr. The info page expects a thumbnail to exist for a scenario, which it should. Adding optional chaining here doesn't make sense given that expectation.

The actual fix is to prevent the deletion of the only scene in a scenario, not working around that problem everywhere else. 

Generic error handling is something we could add, of course, but again I would make that a separate ticket where a comprehensive evaluation is actually performed on the scenario shape.

frontend/src/features/scenarioInfo/components/DetailEditModal.jsx
harbassan: dialog form elements native behaviour, button presses will close the dialog. Our `ModalDialog` component is just a wrapper around native behaviour (it's not actually very good tho, but thats a future concern). Worth adding a comment I suppose.

leowla (approved):

hazikchaudhry (approved):
lgtm.

PR #389, opened 2026-05-09
fix: ScenarioInfo page layout responsiveness

harbassan:
## Issue

The layout of the scenario info page is not responsive, so the "GO" button and description get cut off when the title is too long because everything is sized inconsistently and based on what "roughly" looks right, and scroll is not considered properly with the absolute positioning of the elements. 

## Solution

Changed the styling of the whole page, so that it fits correctly with the design and is also responsive to resizing and longer content.

### before

image

### after

image
image

image

## Risk

No risks, I've tested it on many different realistic (down to 900x500) sizes (i think). 

## Checklist

- [ ] Acceptance criteria met

### DIFF
--- a/frontend/src/features/scenarioInfo/ScenarioInfo.jsx
+++ b/frontend/src/features/scenarioInfo/ScenarioInfo.jsx
@@ -122,7 +122,7 @@ function ScenarioInfo() {
     <div className="bg-base-100 text-base-content">
       {/* Responsive Container optimised for 900x500 min to 1600x900 max */}
       <button
-        className="fixed btn btn-phantom text-m ml-xl mt-l font-dm px-0"
+        className="fixed z-10 btn btn-phantom text-m ml-xl mt-l font-dm px-0"
         onClick={handleBackToPlay}
       >
         <ArrowLeftIcon size={20} />
--- a/frontend/src/features/scenarioInfo/ScenarioInfo.jsx
+++ b/frontend/src/features/scenarioInfo/ScenarioInfo.jsx
@@ -119,61 +119,63 @@ function ScenarioInfo() {
   };
 
   return (
-    <div className="bg-base-100 text-base-content min-h-screen relative overflow-x-hidden">
-      {/* Back Button */}
-      <button
-        className="z-50 bg-transparent border-none text-primary cursor-pointer hover:text-base-content transition-colors pl-xl pt-l font-dm text-s fixed"
-        onClick={handleBackToPlay}
-      >
-        ← Back
-      </button>
-      {/* Responsive Container optimised for 1024x768 min to 1600x900 max */}
-      <div className="min-w-[1024px] max-w-[1600px] mx-auto overflow-hidden">
+    <div className="bg-base-100 text-base-content h-screen overflow-hidden">
+      {/* Responsive Container optimised for 900x500 min to 1600x900 max */}
+      <div className="min-w-[900px] max-w-[1600px] mx-auto flex h-full">
         {/* Sidebar */}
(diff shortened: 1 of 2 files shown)


frontend/src/features/scenarioInfo/ScenarioInfo.jsx
```diff
@@ -121,59 +122,41 @@ function ScenarioInfo() {
...
       </button>
       {/* Responsive Container optimised for 1024x768 min to 1600x900 max */}
-      <div className="min-w-[1024px] max-w-[1600px] mx-auto px-8 lg:px-16 xl:px-24 h-screen flex relative overflow-hidden">
+      <div className="min-w-[1024px] max-w-[1600px] mx-auto overflow-hidden">
         {/* Sidebar */}
-        <div className="w-[27%] bg-base-100 flex flex-col relative h-full overflow-hidden flex-shrink-0">
-          {/* Spacer to push content down */}
-          <div className="h-[35vh] flex-shrink-0"></div>
-
+        <div className="w-[calc(20%+var(--spacing-2xl))] left-[max(0px,calc(50vw-800px))] pl-2xl bg-base-100 flex flex-col fixed h-full overflow-hidden flex-shrink-0 gap-m pt-[calc(var(--spacing-4xl)+var(--spacing-xl)+var(--spacing-l))]">
```
leowla: the many `calc` functions here very hard to understand and uses vars defined elsewhere that could be changed and very easily affect this page by mistake

frontend/src/features/scenarioInfo/ScenarioInfo.jsx
```diff
@@ -121,59 +122,41 @@ function ScenarioInfo() {
     <div className="bg-base-100 text-base-content min-h-screen relative overflow-x-hidden">
       {/* Back Button */}
       <button
-        className="absolute z-50 bg-transparent border-none text-primary cursor-pointer hover:text-base-content transition-colors px-8 py-6 top-0 left-0 font-dm text-s"
+        className="z-50 bg-transparent border-none text-primary cursor-pointer hover:text-base-content transition-colors pl-xl pt-l font-dm text-s fixed"
```
leowla: The back button is not aligned with the sidebar on standard screen sizes like 1920x1080 and 2560x1440

leowla (changes requested):
The many CSS calculations are very hard to understand and look fragile. I have a working version of flexbox layout on [vps-48-review-suggestions](https://github.com/UoaWDCC/VPS/tree/vps-48-review-suggestions) if you want to comapre.

frontend/src/features/scenarioInfo/ScenarioInfo.jsx
harbassan: its not intended to be, based on the figma designs:

image

frontend/src/features/scenarioInfo/ScenarioInfo.jsx
harbassan: those arent just vars defined elsewhere, they are our main layout vars, changing them is intended to change this page (and every page in our application). Thats why im using them here.

harbassan:
@leowla ive added back one of the calcs, which is the padding top calculation for the sidebar. The reason being that according to the design it should be aligned with the scenario metadata, and doing so requires imitating the responsive sizing flow of that section using our utopia spacings.

ive added a comment explaining that one and its just a basic addition so i feel that its ok.

leowla (approved):
A lot more readable now! 👍 

---

> padding top calculation for the sidebar. The reason being that according to the design it should be aligned with the scenario metadata

I personally don't agree with this design choice since it does not account for fluid layouts and in practice will become misaligned with long titles. But this isn't a big deal its just me being picky. What matters is the code is readable.

Image

hazikchaudhry (approved):
looks good to me brochacho

PR #390, opened 2026-05-09
feat(authoring,play): single-scene timer and state operations

leowla:
## Issue

Missing timer feature documented in https://github.com/UoaWDCC/VPS/blob/master/wiki/Specifications/Timer.md

## Solution

Implement the single-scene timer.

- Play: A new timer component that can execute state variable transitions on timeout.
- Authoring: New "Timer Duration" input, "On Timeout" settings dropdown, and "Add Timeout State Operation" modal.

https://github.com/user-attachments/assets/4ced01b5-83ee-4948-9412-cbeeb4b1b34c

image

image

## Risk

- `Select.tsx` input component dependents

## Other

I don't like the settings for the timer is split into scene settings and a separate timeout settings dropdown but I'm pretty limited unless I completely change the dropdown-style ui.

Maybe I could create a mm:ss time input component if inputting seconds is too cumbersome for the user when they want to set longer timers (e.g. 8 minutes) and they don't want to convert minutes to seconds in their head.

We might want a top bar separate from the play area for the timer and resources button so that they don't cover potential clickable areas in the scene.

## Checklist

- [x] Acceptance criteria met

### DIFF
--- a/frontend/src/components/StateVariables/TimerStateOperationMenu.jsx
+++ b/frontend/src/components/StateVariables/TimerStateOperationMenu.jsx
@@ -1,0 +1,217 @@
+import { useContext, useEffect, useState } from "react";
+import { PlusIcon } from "lucide-react";
+import ScenarioContext from "context/ScenarioContext";
+import { getDefaultValue, stateTypes, validOperations } from "./stateTypes";
+import { modifySceneProp } from "../../features/authoring/scene/operations/modifiers";
+import SelectInput from "../../features/authoring/components/Select";
+import ModalDialog from "../ModalDialogue";
+import useVisualScene from "../../features/authoring/stores/visual";
+
+function TimerOperationRow({ operation, index, operations }) {
+  const { stateVariables } = useContext(ScenarioContext);
+
+  const stateVariable = stateVariables?.find(
+    (v) => v.id === operation.stateVariableId
+  );
+
+  const [localOperation, setLocalOperation] = useState(operation.operation);
+  const [localValue, setLocalValue] = useState(operation.value);
+
+  useEffect(() => {
+    setLocalOperation(operation.operation);
+    setLocalValue(operation.value);
+  }, [operation]);
+
+  if (!stateVariable) return null;
+
+  function saveOperation(v) {
+    setLocalOperation(v);
+    const updated = operations.map((op, i) =>
+      i === index ? { ...op, operation: v } : op
+    );
+    modifySceneProp("timerStateOperations", updated);
+  }
+
+  function saveValue(v) {
+    setLocalValue(v);
+    const updated = operations.map((op, i) =>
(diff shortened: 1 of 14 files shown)


frontend/src/features/playScenario/components/SceneTimer.jsx
```diff
@@ -0,0 +1,45 @@
+import { useEffect, useRef, useState } from "react";
+import { TimerIcon } from "lucide-react";
+
+export default function SceneTimer({ duration, onTimeout }) {
+  const [secondsLeft, setSecondsLeft] = useState(duration);
+  const onTimeoutRef = useRef(onTimeout);
+  onTimeoutRef.current = onTimeout;
```
harbassan, line 7: why the ref here?

wiki/Specifications/Timer.md
```diff
@@ -41,8 +41,8 @@ To avoid a situation like this, we can lock the scene(s) to a certain role, disa
 
 When a timer ends, the author probably wants to perform some meaningful action. The most likely "actions" in this case are scene and state transitions. For example:
 
-- *decrement health by 20*
-- *set out_of_time to true*
+- _decrement health by 20_
+- _set out_of_time to true_
```
harbassan, line 45: ?? do u have beef with * syntax

frontend/src/components/StateVariables/TimerStateOperationMenu.jsx
```diff
@@ -0,0 +1,215 @@
...
+          ))}
+        </div>
+      </div>
+
+      <ModalDialog
+        title="Add Timeout State Operation"
+        open={createOpen}
+        onClose={() => setCreateOpen(false)}
+      >
+        {!stateVariables?.length ? (
```
harbassan: id prefer the modal to exist in its own module, given its size and that it has state, for the sake of better organisation of the codebase. ill leave it up to you tho.

harbassan (approved):
sweet, it seems to be working well for me. just a few comments.

frontend/src/features/playScenario/components/SceneTimer.jsx
leowla: For performance so I don't need to add `onTimeout` to the effect's dep array (lines 16-20) and ensure that it still only performs onTimout actions once.

wiki/Specifications/Timer.md
leowla: no but prettier does 🫪

frontend/src/features/playScenario/components/SceneTimer.jsx
harbassan: 👍

hazikchaudhry (approved):
nice works really well, I think ill make a few tickets based on the UI. Otherwise looks really good !

PR #392, opened 2026-05-10
feat: add scene outline overlay

harbassan:
## Issue

When an image or shape extends outside the bounds of the scenes, it obscures the boundary. When something covers all or the majority of the edges, it makes it impossible to see whats actually within the scene and what isn't.

## Solution

Added a simple outline that exists between the overlay and content layers of the canvas, which is always present. This uses a blend mode to make sure it is always visible regardless of the content on the scene.

image

## Risk

None.

## Checklist

- [x] Acceptance criteria met

### DIFF
--- a/frontend/src/features/authoring/canvas/Canvas.tsx
+++ b/frontend/src/features/authoring/canvas/Canvas.tsx
@@ -80,6 +80,25 @@ function Canvas() {
         onContextMenu={handleContextMenu}
       >
         <Overlay />
+
+        {/* scene outline */}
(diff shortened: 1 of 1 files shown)


del-ereno (approved):
code looks good - the line is a little hard to see sometimes. ~~Google slides has a cool approach where a red outline appears whenever an object is crossing/over the outline~~ nvm this looks like a pain to implement. this clean gjgj

leowla (approved):

PR #393, opened 2026-05-11
Vps 9/conditions in collections

rgin216:
## Issue

Resource collections could not have their own state conditions. Authors could only conditionally reveal individual files, which made it awkward to reveal a whole collection of related resources at once. The old CollectionChild model was also unused.

## Solution

Added stateConditionals to collection groups, with create/update/delete API support. Generalized the existing state conditional UI so it can target either a file or a collection. Player resource filtering now requires both collection-level and file-level conditions to pass before a file is visible.

Also removed the unused CollectionChild model after confirming it had no references.

## Risk

Collection-level conditions add another visibility gate, so incorrectly configured collection conditions can hide files that would previously have appeared based only on file-level conditions. Existing collections without conditions should continue to behave as before.

## Checklist

- [ ] Acceptance criteria met

## Summary by CodeRabbit

* **New Features**
  * Add group-level state conditions and display them alongside file-level conditions.
  * Unified create/edit/delete flows for state conditions that work for both files and groups, with a selected-target header and configurable endpoints.
  * Groups are excluded from views when their group-level conditions are not met.

* **Bug Fixes**
  * Fixed file response serialization to remove an internal storage ID property during condition operations.


GogaIV (approved):
Lgtm 👌

frontend/src/components/StateVariables/StateConditionalMenu.jsx
```diff
@@ -8,42 +8,61 @@ import EditStateConditional from "./EditStateConditional";
...
  */
-const StateConditionalMenu = ({ file, updateFile }) => {
-  if (!file) {
+const StateConditionalMenu = ({
+  target,
+  title = "State Conditionals",
+  endpoint,
+  updateTarget,
+  file,
+  updateFile,
```
harbassan: not actually passed in by the parent, redundant.

harbassan (approved):
awesome, i've tested it with a few different conditions + conditions on the children and its working well. just the minor thing mentioned.

PR #396, opened 2026-05-16
feat(authoring): start play from the currently viewed scene

leowla:
## Issue

Authors can not easily restart their scenario play-test.

## Solution

Clicking "Play" will start a play session from the currently viewed scene in the editor.

https://github.com/user-attachments/assets/0dea7701-c9e5-405a-a561-6563fe6a5e9e

## Technical

- Play button passes current scene as ?startScene query param
- Resolver preserves the query param through the singleplayer/multiplayer redirect
- PlayScenarioPage sends startScene to the navigate API on first load only
- Navigate endpoint verifies the user is an author before honouring startScene; regular players cannot use it to jump to arbitrary scenes
- Resets the user's path to the requested scene so forward navigation stays consistent

## Risk

- The startScene override is gated behind `isAuthor` but I have not thoroughly tested if actually blocks regular players.
- Currently no option to edit state variables that would have been changed by previous scenes

## Checklist

- [x] Acceptance criteria met

### DIFF
--- a/backend/src/db/daos/accessDao.js
+++ b/backend/src/db/daos/accessDao.js
@@ -105,10 +105,17 @@ const revokeAccess = async (scenarioId, userId) => {
   return { status: 200, message: "Revoked" };
 };
 
+const isAuthor = async (scenarioId, uid) => {
+  const accessList = await getAccessList(scenarioId);
+  if (!accessList) return false;
+  return accessList.ownerId === uid || accessList.users.has(uid);
+};
+
 export {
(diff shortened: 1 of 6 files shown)


harbassan (approved):
awesome I've tested it and it seems to work well, nice handling of the concurrent navigation edge case as well. 😁 🥳

glybrio (approved):
nice - just have to open a new issue for state variables but this is awesome for testing the branching!

PR #399, opened 2026-05-19
Vps 91/prevent last scene deletion

rgin216:
## Issue

Deleting the only scene in a scenario allowed the scenario to end up with no scenes. This caused the authoring page and scenario info page to crash because both reasonably expect every scenario to have at least one scene.

## Solution

Scenarios are now created with an initial Scene 1 on the backend, making the “at least one scene” invariant true from creation.

The scene delete API now rejects attempts to delete the last remaining scene with a 409 response. The reorder API also rejects empty sceneId arrays, so a scenario cannot be reordered into having no scenes.

The authoring UI now disables the delete action when only one scene remains and restores the scene list if a delete request fails.

## Risk

Low. The main behaviour change is that scenarios can no longer be empty. Any existing code or tests that expected newly created scenarios to have zero scenes may need updating.

Existing database records with no scenes are not automatically backfilled by this PR.

## Checklist

- [ ] Acceptance criteria met


backend/src/routes/api/scene.js
```diff
@@ -90,6 +92,12 @@ router.put("/reorder", async (req, res) => {
     return res.status(400).json({ error: "sceneIds must be an array" });
   }
 
+  if (sceneIds.length === 0) {
+    return res.status(status.CONFLICT).json({
+      error: "A scenario must have at least one scene.",
+    });
+  }
+
```
harbassan: given that this route is about reordering, this specific gate doesn't make much sense. A better gate would probably be a check that the number of scenes is the same as it was before, to make sure this route is single purpose only.

backend/src/routes/api/scene.js
```diff
@@ -13,11 +13,13 @@ import {
 import auth from "../../middleware/firebaseAuth.js";
 import scenarioAuth from "../../middleware/scenarioAuth.js";
 import validScenarioId from "../../middleware/validScenarioId.js";
+import status from "../../util/status.js";
 
 const router = Router({ mergeParams: true });
 
 const HTTP_OK = 200;
 const HTTP_NOT_FOUND = 404;
+const LAST_SCENE_DELETE_MESSAGE = "A scenario must have at least one scene.";
```
harbassan: if this is only going to be in one place, there's no point in making it a constant here.

backend/src/routes/api/scene.js
```diff
@@ -90,6 +92,12 @@ router.put("/reorder", async (req, res) => {
     return res.status(400).json({ error: "sceneIds must be an array" });
   }
 
+  if (sceneIds.length === 0) {
+    return res.status(status.CONFLICT).json({
+      error: "A scenario must have at least one scene.",
+    });
```
leowla:
```suggestion
    return res
      .status(status.CONFLICT)
      .json({ error: LAST_SCENE_DELETE_MESSAGE });
```
reuse your variable for future proofing

leowla (approved):
🎬🗑️

harbassan (approved):
awesome 😃

PR #401, opened 2026-05-19
Fixed State Conditional Resolution

RLee64:
## Issue
Problem with evaluating state conditionals for boolean state variables. Results are often incorrect (i.e. for '=', resource never shows). Numbers are also stored as strings, resulting in a need to parse before comparison.

## Solution
Allow for mixed data types for state variables. Now also compares boolean variables directly rather than converting from string format. Numbers properly stored as numbers (for both creating and editing state conditionals). Comparison logic simplified as type-dependent parsing is no longer necessary.

## Risk
None that I'm aware of.

## Checklist

- [ ] Acceptance criteria met

### DIFF
--- a/backend/src/db/models/StoredFile.js
+++ b/backend/src/db/models/StoredFile.js
@@ -29,7 +29,7 @@ const StoredFileSchema = new Schema(
             enum: ["=", "!=", "<", ">"],
             required: true,
           },
-          value: { type: String, required: true },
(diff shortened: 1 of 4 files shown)


harbassan (approved):
yeah awesome fam 👍 😀 👍 just fix the lint error please

hazikchaudhry (approved):
looks good bro !

PR #402, opened 2026-05-19
fix: save null value scene links

harbassan:
## Issue

Setting a scene link to **None** doesn't save, making unsetting a link impossible.

## Solution

since null is typeof object, the modification function is trying to perform an object merge after the check. Obviously this property is not an object, so the solution is just to explicity check for null, and if so treat it as direct assignment.

## Risk

No risk.

## Checklist

- [x] Acceptance criteria met


glybrio (approved):
one line edits can't go wrong right? 

THANKS GOAT

br-Chan (approved):
> In the beginning ~~the Universe was created~~ null was made typeof object. This has made a lot of people very angry and been widely regarded as a bad move.

~ Douglas Adams

LGTM!

PR #403, opened 2026-05-19
Vps 16 add keyboards shortcuts to the editor

FridgeProtector1:
## Issue

The editor needs shortcut handling to respect OS-specific modifier keys so the same shortcut system works correctly on both Mac and Windows.

## Solution

Add a centralized keyboard shortcut handling layer that listens for key events and maps them to existing editor actions. The shortcut logic uses a shared shortcut definition structure to match key combinations against the current platform’s primary modifier key, so Cmd/metaKey is used on Mac and Ctrl is used on Windows. This means Ctrl is not treated as the same shortcut on macOS unless it is explicitly bound that way.

When a shortcut is matched, the handler dispatches to the same existing functions used by the editor UI rather than reimplementing the underlying behavior. 

## Risk

Not too sure but perhaps there could be conflicts with browser specific shortcuts (Didn't notice anything but idk any browser short cuts anyways). 
## Shortcuts mapped
On Windows/Linux, mod means Ctrl
On macOS, mod means Cmd

Z index commands don't work too well but I believe they weren't working to begin with, but they are mapped too.

Shortcuts
mod+z → Undo
mod+shift+z → Redo
mod+y → Redo
mod+d → Duplicate selected component
Backspace / Delete → Remove selected component
only when not in text mode
mod+ArrowUp → Bring forward one layer
mod+Shift+ArrowUp → Bring to front
mod+ArrowDown → Send backward one layer
mod+Shift+ArrowDown → Send to back
Text mode shortcuts
mod+a → Select all text
mod+b → Toggle bold
mod+i → Toggle italic
mod+u → Toggle underline
Existing non-shortcut keyboard behavior still kept
Arrow keys move selected components by 5px when not in text mode
In text mode, arrow keys / Home / End / typing / Enter / Backspace still use the existing text editor behavior
Platform note

## How to add more
Add new shortcuts in frontend/src/features/authoring/handlers/keyboard/shortcuts.ts by defining a new key combination and mapping it to an existing editor action. keyboard.ts already routes global keydown events through handleShortcut(e), so new bindings only need to be added to the shortcut map and matched with the platform-specific primary modifier.

## Checklist

- [ ] Acceptance criteria met

## Summary by CodeRabbit

* **New Features**
  * Expanded editor keyboard shortcuts for undo/redo, duplicate, delete, layer ordering, and text editing (select all and bold/italic/underline toggles).
  * Improved cross-platform shortcut detection for Mac-like devices (including iOS/iPadOS).

* **Bug Fixes**
  * Prevented shortcuts from interfering with typing in editable fields.
  * Updated text-mode key routing so Arrow/Home/End consistently navigate, while other keys edit.
  * Improved text styling consistency for selected vs partial selections.
  * Made component duplication fail more predictably when the source is missing.


frontend/src/features/authoring/handlers/keyboard/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export function isMacPlatform() {
+  const platform = getPlatform();
+  return MAC_PLATFORMS.some((name) => platform.includes(name));
+}
+
+export function isPrimaryShortcutModifier(e: KeyboardEvent) {
+  return isMacPlatform() ? e.metaKey : e.ctrlKey;
+}
+
```
leowla, line 22:
```suggestion
export function isPrimaryShortcutModifier(e: KeyboardEvent) {
  return isMacPlatform() ? e.metaKey : e.ctrlKey;
}

export function isSecondaryShortcutModifier(e: KeyboardEvent) {
  return isMacPlatform() ? e.ctrlKey : e.metaKey;
}

```

frontend/src/features/authoring/handlers/keyboard/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+function hasOnlyRequestedModifiers(e: KeyboardEvent, modifiers: Set<string>) {
+  const usesPrimary = modifiers.has("mod");
+  const usesCtrl = modifiers.has("ctrl");
+  const usesMeta = modifiers.has("meta");
+  const usesShift = modifiers.has("shift");
+  const usesAlt = modifiers.has("alt") || modifiers.has("option");
+
+  if (usesPrimary) {
+    if (!isPrimaryShortcutModifier(e)) return false;
```
leowla, line 48:
```suggestion
    if (!isPrimaryShortcutModifier(e)) return false;
    if (isSecondaryShortcutModifier(e)) return false;
```
Also check the non-primary key is not held.
e.g. Cmd+Ctrl+Z on Mac and Ctrl+Win+Z on Windows should not activate undo.

frontend/src/features/authoring/handlers/keyboard/utils.ts
```diff
@@ -0,0 +1,67 @@
+const MAC_PLATFORMS = ["mac", "iphone", "ipad", "ipod"];
+
+function getPlatform() {
+  return (
+    globalThis.navigator?.platform ??
+    globalThis.navigator?.userAgent ??
+    ""
+  ).toLowerCase();
```
leowla:
```suggestion
  const nav = globalThis.navigator;
  return (nav?.platform || nav?.userAgent || "").toLowerCase();
```
navigator.platform can return empty string instead of null so || is safer than ?? to catch that

frontend/src/features/authoring/handlers/keyboard/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export function isMacPlatform() {
+  const platform = getPlatform();
+  return MAC_PLATFORMS.some((name) => platform.includes(name));
+}
```
leowla, line 17:
```suggestion
let cachedIsMac: boolean | undefined;

export function isMacPlatform() {
  if (cachedIsMac === undefined) {
    const platform = getPlatform();
    cachedIsMac = MAC_PLATFORMS.some((name) => platform.includes(name));
  }
  return cachedIsMac;
}
```

We don't need to check if is Mac for every keyboard press, once is enough.

leowla (changes requested):
Looks good so far! My main issue is with checking the non-primary key is not held. The other stuff is minor fixes that you can ignore if I'm wrong.

harbassan (approved):
finally 🎊

hazikchaudhry (approved):
it works, lgtm

leowla (approved):

PR #406, opened 2026-05-20
feat: audio playback on scenario

harbassan:
## Issue

Very basic audio implementation for the PlayScenarioPage does not exist.

## Solution

The audios attached to the scene will now play, once the user has enabled audio for the scenario. This explicit enable is required due to browser laws, which are there to protect the user from unwanted playback. After the enable the rest of the scenario will autoplay audios.

## Risk

I did this very hastily and just did basic testing. It does work but it is probably not the most efficient way to do this.

## Checklist

- [x] Acceptance criteria met

### DIFF
--- a/frontend/src/features/playScenario/PlayScenarioPage.jsx
+++ b/frontend/src/features/playScenario/PlayScenarioPage.jsx
@@ -85,7 +85,7 @@ const navigateMultiplayer = async (
 };
 
 function playAudios(scene) {
-  const audios = scene.components.filter(c => c.type === "audio");
+  const audios = scene.components.filter((c) => c.type === "audio");
   const playables = [];
   for (const audio of audios) {
     const playable = new Audio(audio.url);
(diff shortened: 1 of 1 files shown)


hazikchaudhry (approved):
looks good 👨‍💻

frontend/src/features/playScenario/PlayScenarioPage.jsx
```diff
@@ -305,6 +334,14 @@ export default function PlayScenarioPage({ group }) {
           />
         </div>
       )}
+      {!audioAllowed && (
+        <div className="absolute top-4 left-4 z-30">
```
br-Chan, line 338: Not familiar with this part of the code, why is there a `z-30` here?

br-Chan:
> I did this very hastily and just did basic testing. It does work but it is probably not the most efficient way to do this.

I just read this, haha. Sounds good, and I suppose if you want to come back later and improve it then just make a note of it so we don't forget, otherwise LGTM

frontend/src/features/playScenario/PlayScenarioPage.jsx
hazikchaudhry: i think to make sure its always on top when it comes to elements on the screen, but yea not the best way. I approved incase we want this for the milestone

glybrio (approved):
good for demo today

PR #407, opened 2026-05-20
test: add test helpers and fix suite bugs

leowla:
## Issue

- the Express error handler had wrong arity (2 params instead of 4),
- nextFunction mock was never cleared between tests causing cross-test contamination
- deprecated Mongoose connection options were passed
- the DELETE scenario test had no assertion that the access list entry was removed.
- user.js had a pre-existing bug where Object.entries was called on a Mongoose query result array, passing whole documents as IDs instead of extracting ._id

## Solution

Fix bugs and dedupe Mongoose connection code in test suite

## Risk

Low. We can rely on the test suite!!!
Potential risk with user.js fix which corrects broken behaviour (user assignment was passing document objects instead of IDs to the query)

## Checklist

- [x] Acceptance criteria met


harbassan (approved):
awesome, that should make implementing testing much much easier 😄. approving for the sake of progress on this, but I think it might be a good idea to rename `mongoSetup.js` given that it also has the express setup helper within it.

K1mmyn (approved):
lgtm maintains all functionality and you improved the code overall

nice docs

PR #409, opened 2026-05-21
Vps 40/dropdown trunucation

GogaIV:
## Issue

drop down truncation 

## Solution

it was a little small i felt you could only fit 22 chars so i expanded it and added a scroll for overflow
I also added an extra quick command for lint and prettier on the side cause it saves a few seconds

## Images
image

## Risk

None

## Checklist

- [ ] Acceptance criteria met

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -11,7 +11,9 @@
 		"lint": "eslint src",
 		"lint:fix": "eslint --fix src",
 		"prettier": "prettier --check src",
-		"prettier:write": "prettier --write src"
+    "prettier:write": "prettier --write src",
+    "prettierlint": "prettier --check src && eslint src",
+    "prettierlint:fix": "prettier --write src && eslint --fix src"
 	},
 	"dependencies": {
 		"axios": "^1.7.7",
--- a/backend/package.json
+++ b/backend/package.json
@@ -11,7 +11,7 @@
 		"lint": "eslint src",
 		"lint:fix": "eslint --fix src",
 		"prettier": "prettier --check src",
-    "prettier:write": "prettier --write src"
+		"prettier:write": "prettier --write src"
 	},
 	"dependencies": {
 		"axios": "^1.7.7",
--- a/backend/package.json
+++ b/backend/package.json
@@ -11,9 +11,7 @@
 		"lint": "eslint src",
 		"lint:fix": "eslint --fix src",
 		"prettier": "prettier --check src",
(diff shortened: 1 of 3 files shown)


leowla (changes requested):
It would be better to wrap text here.

e.g.
Image

https://primer.style/accessibility/patterns/truncation/#interactive-elements

> Do not truncate text if it contains focusable elements such as links, buttons, and mentions/tags.
>
>If you experience this challenge, some potential design alternatives are:
> - Wrap text instead of truncate (so long as focus does not break)
> - Use a tooltip that is keyboard and screen reader accessible to announce the entirety of text
> - Offer a method such as a button or toggle to "expand all titles" and "hide titles"

GogaIV:
image

sum like this

backend/package.json
```diff
@@ -11,7 +11,9 @@
 		"lint": "eslint src",
 		"lint:fix": "eslint --fix src",
 		"prettier": "prettier --check src",
-		"prettier:write": "prettier --write src"
+    "prettier:write": "prettier --write src",
+    "prettierlint": "prettier --check src && eslint src",
+    "prettierlint:fix": "prettier --write src && eslint --fix src"
```
leowla: These are not related to the dropdown change

leowla (changes requested):
You can do `yarn run prettier lint` to save time. We could set up a pre-commit hook to properly solve this problem but anyway this should be done in a separate PR since its not related to the dropdown.

backend/package.json
GogaIV: didnt know i could do that lowk

backend/package.json
```diff
@@ -11,7 +11,7 @@
 		"lint": "eslint src",
 		"lint:fix": "eslint --fix src",
 		"prettier": "prettier --check src",
-		"prettier:write": "prettier --write src"
+    "prettier:write": "prettier --write src"
```
leowla:
```suggestion
		"prettier:write": "prettier --write src"
```

Sorry I'm nitpicky about this since the rest of the file uses tabs

leowla (approved):
📐

backend/package.json
GogaIV: all good, i saw it but i was confused what the diff was didn't realize it was probably my tab space sizing

harbassan (approved):
beautiful 😀 🌻

PR #411, opened 2026-05-25
Vps 83/new UI delete scenerios

rgin216:
## Issue

There's no way to delete a scenario in the newer UI

## Solution

Added a small delete button on each scneario in Create&Edit
image

## Risk

Should have no issues

## Checklist

- [ ] Acceptance criteria met

## Summary by CodeRabbit

* **New Features**
  * Added an in-page “Delete Scenario” confirmation dialog to require explicit confirmation before removal.
* **Accessibility Improvements**
  * Improved scenario delete controls with clearer labels (aria-label/title) and more reliable click handling.
* **Bug Fixes**
  * Updated deletion flow to use a pending, in-page confirmation state rather than immediate deletion.
* **Style**
  * Cleaned up minor button styling spacing in the notes list.


harbassan (changes requested):
functionally perfect, ill leave it to @del-ereno for his UI input, by I think it looks good. Just need to demuddle this pr from your other changes from your previous pr. (just cherry pick it)

del-ereno:
yeah functionality is great. Can be a little annoying to click when the card moves up on mouse hover. 

One thing I'm little confused on, not from your changes, is the edit scenario details page only being accessible from the 'play' page? 'play' shows every scenario the user has play access to, and 'play&edit' only shows the ones where user has edit perms. Would it not make more sense for this to be in 'create&edit', and delete to sit inside or alongside it?  
image

harbassan:
> yeah functionality is great. Can be a little annoying to click when the card moves up on mouse hover.
> 
> One thing I'm little confused on, not from your changes, is the edit scenario details page only being accessible from the 'play' page? 'play' shows every scenario the user has play access to, and 'play&edit' only shows the ones where user has edit perms. Would it not make more sense for this to be in 'create&edit', and delete to sit inside or alongside it? image

yep this is an inconsistency i keep noticing every so often and then forgetting about, we definitely need to change this so that the "scenario info" page also shows up in create and edit, and the edit only exists there. Ill make a ticket for it

harbassan (approved):
modified slightly to use the modal compoennt we already have and to make the hover remain consistent with the rest of the app.

PR #413, opened 2026-05-26
VPS-96/Allowed renaming of resources in edit view

kmck133:
## Issue

After uploading a resource, there was no way to rename it. Users were stuck with the original filename even if it was unclear or incorrect.

## Solution

Added the ability to rename uploaded resources from the collections sidebar in the manage resources view. A pencil icon appears next to each file; clicking it replaces the filename with an inline input field. The user can confirm with Enter or the checkmark button, or cancel with Escape or the X button.
This required a new PATCH /api/files/:fileId endpoint on the backend that validates and updates the name field on StoredFile. The frontend calls this endpoint, then updates both the sidebar list and the active preview panel (if the renamed file is currently selected) without a page reload.

## Risk

- Low. Renaming a resource updates the display name in StoredFile but not the filename stored in GridFS. This is intentional, as the application always resolves filenames through StoredFile.name (including when downloading the resource), so there is no functional impact. The divergence is only visible when inspecting the raw resources.files collection in MongoDB directly.
- The new PATCH /:fileId route sits alongside the existing DELETE /:fileId route with no ambiguity since they use different HTTP methods.

## Summary by CodeRabbit

* **New Features**
  * Added inline file renaming in the Resources management UI with Save/Cancel controls and keyboard shortcuts (Enter to save, Escape to cancel), including auto-focus and trimmed names.
* **Usability**
  * Added filename validation (max length) and prevents no-op renames.
  * Improved layout/styling for clearer file selection and download/preview alignment.
* **Reliability**
  * Enhanced rename update flow with appropriate “not found” and error handling when changes can’t be saved.


leowla (changes requested):
There should be a character limit

Image

frontend/src/features/resources/ManageResourcesPage.jsx
```diff
@@ -133,6 +137,55 @@ export default function ManageResourcesPage() {
     }
   }
 
+  async function renameFile(fileId, newName) {
+    try {
+      const user = getAuth().currentUser;
+      if (!user) {
+        toast.error("You must be logged in.");
+        return;
+      }
```
harbassan: this check is unecessary, given that this is on a protected route, so we can make the assumption that they are logged in.

harbassan (changes requested):
just this one + the coderabbit stuff which is valid.

PR #417, opened 2026-05-26
Vps 31/number input objects

GogaIV:
## Issue

Need num input for x y width and height for canvas objects

## Solution

Use the same method as when resizing objects with the mouse, add num input for x y and then calc width based on both x and height based on both y's,
Made speech box and line include property clickable so that they can be registered by the property editor 

## Vid

https://github.com/user-attachments/assets/3d542e90-b03d-493c-bce4-4fa6664d5e5d

## Risk

none isolated component
If there was a reason that speech box and line didnt have property clickable there is a different method I can use to make them editable via the new object property editor

## Checklist

- [ ] Acceptance criteria met

## Summary by CodeRabbit

* **New Features**
  * Added an **Object Properties** panel in the sidebar with numeric fields to edit an object’s **position (X/Y)** and **size (width/height)**.
  * Sidebar now shows property controls for the **currently selected** canvas object.

* **Enhancements**
  * Updated **line** and **speech** component types to be **clickable by default**.


br-Chan:
In regards to input validation, could we instead prevent non-number inputs in each field rather than allowing the user to type characters?

GogaIV:
> In regards to input validation, could we instead prevent non-number inputs in each field rather than allowing the user to type characters?

yeah that was an option between like 3 other options I had 

https://github.com/user-attachments/assets/4355d24c-ecdf-40be-b3c2-9b159813d605

i was inputting letters it was jst rejecting 

looked like this, it was okay  but i couldn't decide so i just copied google slides implementation it jst tells if you input something incorrectly

I can do either idm one just saves val prior to checking if its a num other checks first then saves

frontend/package.json
```diff
@@ -80,5 +80,5 @@
       "last 1 safari version"
     ]
   },
-  "proxy": "http://localhost:5001"
+  "proxy": "http://localhost:5002"
```
leowla:
```suggestion
  "proxy": "http://localhost:5001"
```
revert

leowla (commented):
Does not work for line and speech box. Is this intentional?

GogaIV:
> Does not work for line and speech box. Is this intentional?

No I didn't test it but I also assumed it would cuz it should all be component objects, let me look into it

leowla:
You can update your branch with master since a lot has changed and your PR might be affected

PR #421, opened 2026-06-03
VPS-95/Fixed issue where pasting into input area would paste in slide

kmck133:
## Issue

Pasting text into the scene name input field incorrectly pastes into the slide instead.

## Solution

Added an isInputTarget guard to the global copy, cut, and paste clipboard handlers so they bail out early when the event target is an INPUT or TEXTAREA element, matching the same guard already in the keydown handler.

## Risk

Low. The change only affects clipboard events originating from native input elements, which the global handlers should never have been handling.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed clipboard operations (copy, cut, paste) interfering with form field interactions. Clipboard actions now work correctly when users interact with text input fields without triggering unintended editor behavior.

### DIFF
--- a/frontend/src/features/authoring/handlers/keyboard/clipboard.ts
+++ b/frontend/src/features/authoring/handlers/keyboard/clipboard.ts
@@ -23,7 +23,13 @@ function plainToDoc(text: string) {
   return { style: {}, blocks };
 }
 
+function isInputTarget(e: ClipboardEvent) {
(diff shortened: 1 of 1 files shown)


harbassan (approved):
yep works, clean implementation, nice 👍 😺

leowla (approved):
🐛

PR #423, opened 2026-06-05
Vps 109/add group member from UI

rgin216:
## Issue

Managing scenario group members required exporting the group CSV, manually adding a row, and importing it back into the application.

## Solution

Added a UI flow on the manage-groups page to add a single member directly from the application. This includes a modal for name, email, role, and group number, plus a backend endpoint that appends the member without rebuilding all groups. The endpoint validates required fields, duplicate emails, and duplicate roles within a group.

## Risk

Low to moderate. This adds a new group mutation path, so the main risk is inconsistent group/role data if validation misses an edge case. Existing CSV import behaviour is unchanged.

## Checklist

- [ ] Acceptance criteria met

## Summary by CodeRabbit

* **New Features**
  * Added "Add Member" modal and button to add group members (name, email, role, group number) with trimming/validation and submit state.

* **Bug Fixes**
  * Backend now robustly handles concurrent member additions and legacy group records to prevent duplicates and ensure members are added reliably.
  * Role/email duplicate checks are case/whitespace-insensitive.

* **Tests**
  * Expanded API tests, including a concurrency case verifying single-group creation with multiple simultaneous adds.

### DIFF
--- a/backend/src/db/models/group.js
+++ b/backend/src/db/models/group.js
@@ -16,6 +16,9 @@ const groupSchema = new Schema({
   scenarioId: {
     type: String,
   },
+  group: {
+    type: String,
+  },
   currentFlags: [String],
   stateVariables: [Schema.Types.Mixed],
   stateVersion: {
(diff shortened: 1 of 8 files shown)


backend/src/db/models/group.js
```diff
@@ -16,6 +16,9 @@ const groupSchema = new Schema({
   scenarioId: {
     type: String,
   },
+  group: {
+    type: String,
+  },
```
harbassan, line 21: i dont quite understand this field and the logic associated with it . what does the "group" of the group actually mean?

rgin216:
Changed to be in dashboard not in a scenario

PR #426, opened 2026-06-19
fix(frontend): detangle saving and scene switching logic

harbassan:
## Issue

The scene saving dispatch logic, especially the scene switching, is tangled throughout the history, scene, operation and scene context parts of the code base. The result is scene switching and saving behavior that feels sluggish to use.

## Solution

Scene switching logic has been moved to a central place: SceneContext, and it is only called from one place: AuthoringTool. Instead of class-like initialisation of the history module that was required to allow the history functions to perform actions, the history module is now just a store of the history undo/redo stack that dispatches events when those functions are called. The actual actioning of the undo and redo is performed in the same place as the logic it requires, making use of the SceneContext.

### Important!

MANUAL SAVE is mostly removed in this PR, in favour of a modification-aware autosave. With the addition of the history events, the autosave now just listens for any of these events to be dispatched as a sign of "modification". To prevent against server overload, there is a 2000ms debounce.

I didn't initially intend to change this behaviour, but it became a natural extension using the history events system.

There is a NOTE in the code about the combining of the save functionality and the "history actioning" functionality in the AuthoringTool component. **This is not ideal**, but i've done it this way due to the limitation of SceneContext only being accessible from within the React tree. If we move the scene context into a zustand store, this should be separated for better structuring of the codebase.

## Risk

Both saving and history are modified in this PR, so its actually quite risky. Need to test well.

## Checklist

- [x] Acceptance criteria met

## Summary by CodeRabbit

* **Refactor**
  * Unified scene persistence around a direct “modify” workflow with optimistic updates and error rollback.
  * Reworked undo/redo into a purely event-driven history model that restores editor state via update events.
  * Simplified scene switching and component/state restoration to keep the authoring UI and persisted scene in sync.
  * Updated supporting authoring flows (scene name changes, sidebar scene switching, state variable edits) to use the new modify workflow.
* **Chores**
  * Added typed event/history support (including a small dependency) and introduced a reusable typed debounce helper.

### DIFF
--- a/frontend/src/context/SceneContextProvider.jsx
+++ b/frontend/src/context/SceneContextProvider.jsx
@@ -1,14 +1,15 @@
-import { useCallback, useContext, useEffect } from "react";
+import { useContext } from "react";
 import AuthenticationContext from "./AuthenticationContext";
 import SceneContext from "./SceneContext";
-import { useParams } from "react-router-dom";
+import { useParams, useHistory } from "react-router-dom";
 import { api } from "../util/api";
 import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
 import LoadingPage from "../features/status/LoadingPage";
 import GenericErrorPage from "../features/status/GenericErrorPage";
 import toast from "react-hot-toast";
 import { parseMedia } from "../firebase/storage";
-import { init } from "../features/authoring/scene/history";
+import useEditorStore from "../features/authoring/stores/editor";
+import { replace } from "../features/authoring/scene/operations/modifiers";
 
 async function getAllScenes(user, id) {
   const res = await api.get(user, `api/scenario/${id}/scene/all`);
@@ -23,7 +24,7 @@ function deleteScene(user, scenarioId, sceneId) {
   return api.delete(user, `/api/scenario/${scenarioId}/scene/${sceneId}`);
 }
 
-async function saveScenePatch(user, scenarioId, patch) {
+async function modifyScene(user, scenarioId, patch) {
   const parsedComponents = await parseMedia(
     patch.components,
     scenarioId,
@@ -37,6 +38,62 @@ async function saveScenePatch(user, scenarioId, patch) {
   });
 }
 
+function generatePatch(modified, saved) {
+  const components = [];
+  const deletedComponentIds = [];
+  const fields = {};
+
+  const currentComponents = modified.components ?? {};
+  const savedComponents = saved.components ?? [];
+
+  Object.entries(currentComponents).forEach(([id, component]) => {
+    if (
+      JSON.stringify(component) !==
+      JSON.stringify(savedComponents.find((c) => c.id === id))
+    ) {
+      components.push(structuredClone(component));
+    }
+  });
+
+  savedComponents.forEach((c) => {
+    if (!currentComponents[c.id]) deletedComponentIds.push(c.id);
+  });
+
+  ["name", "roles", "time", "directLink", "timerStateOperations"].forEach(
+    (field) => {
+      if (JSON.stringify(modified[field]) !== JSON.stringify(saved[field])) {
+        fields[field] = structuredClone(modified[field]);
+      }
+    }
+  );
+
+  return {
+    _id: modified._id,
+    fields,
+    components,
+    deletedComponentIds,
+  };
+}
+
+function applyPatch(scene, patch) {
+  const { fields = {}, components = [], deletedComponentIds = [] } = patch;
+
+  const updated = [];
+  const seen = new Set();
+
+  for (const c of scene.components) {
+    if (deletedComponentIds.includes(c.id)) continue;
+    updated.push(components.find((uc) => uc.id === c.id) ?? c);
+    seen.add(c.id);
+  }
+
+  for (const c of components) {
+    if (!seen.has(c.id)) updated.push(c);
+  }
+
+  return { ...scene, ...fields, components: updated };
+}
+
 /**
  * This is a Context Provider made with the React Context API
  * SceneContextProvider allows access to scene info and the refetch function
@@ -44,6 +101,7 @@ async function saveScenePatch(user, scenarioId, patch) {
 export default function SceneContextProvider({ children }) {
   const { user } = useContext(AuthenticationContext);
   const { scenarioId } = useParams();
+  const history = useHistory();
 
   const queryClient = useQueryClient();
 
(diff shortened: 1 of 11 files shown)


frontend/src/context/SceneContextProvider.jsx
```diff
@@ -95,25 +153,50 @@ export default function SceneContextProvider({ children }) {
...
-      init(scenesQuery.data, scenarioId, saveScenePatchWrapper);
-    }
-  }, [scenesQuery.data, scenarioId, saveScenePatchWrapper]);
+  async function switchScene(scene, id) {
+    await modifyMutationWrapper(scene);
+    useEditorStore.getState().clear();
+    replace(scenesQuery.data.find((s) => s._id === id));
+    history.push({ pathname: `/scenario/${scenarioId}/scene/${id}` });
+    localStorage.setItem(`${scenarioId}:activeScene`, id);
+  }
```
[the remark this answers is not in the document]
harbassan: this is on purpose since its optimisic UI

frontend/src/context/SceneContextProvider.jsx
```diff
@@ -95,25 +153,58 @@ export default function SceneContextProvider({ children }) {
...
 
-  const saveScenePatchWrapper = useCallback(
-    (patch) => saveScenePatchMutation.mutateAsync(patch),
-    [saveScenePatchMutation.mutateAsync]
-  );
-
-  useEffect(() => {
-    if (scenesQuery.data && scenarioId) {
-      init(scenesQuery.data, scenarioId, saveScenePatchWrapper);
+  function modifyMutationWrapper(scene) {
```
rgin216: `modifyMutationWrapper` uses `modifyMutation.mutate(...)`, which is fire-and-forget and does not return the PATCH promise. But `switchScene` awaits `modifyMutationWrapper`, and other callers also await `modifyScene(...)`, so the code currently continues before the save/upload has actually completed.

I think this should use `mutateAsync` and return the promise, or callers should stop treating it as awaited.

frontend/src/features/authoring/AuthoringToolPage.jsx
```diff
@@ -24,28 +26,44 @@ const listeners = [
...
 
-  // autosave setup
+  // NOTE: this is both the autosaver and the history actioner, which are distinct
+  // operations, but are both here due to the limitations of the scene context
   useEffect(() => {
-    if (!sceneId) return;
-    const autosave = setInterval(save, AUTOSAVE_INTERVAL);
-    return () => clearInterval(autosave);
-  }, [sceneId]);
+    const debounced = debounce(save, 5000);
```
rgin216: This effect installs the history/autosave listener with an empty dependency array, so it captures the initial values of `sceneId`, `saving`, `modifyScene`, `switchScene`, and `save`. That makes the `saving` guard unreliable and can also make cross-scene undo/redo compare against a stale scene id.

I think this listener should either include correct dependencies or use refs for values that need to stay current inside the event handler/debounced save.

frontend/src/features/authoring/AuthoringToolPage.jsx
```diff
@@ -24,28 +26,44 @@ const listeners = [
...
+
+    const listener = async ({ operation, record }) => {
+      if (operation === "undo" || operation === "redo") {
+        if (record.sceneId !== sceneId) switchScene(getScene(), record.sceneId);
+        const state = operation === "undo" ? record.before : record.after;
+        if (state === null) setSelected(null);
+        replaceComponent(record.id, state);
```
rgin216, line 65: When `record.state === null`, this clears selection and removes the component, but then immediately calls `setSelected(record.id)`. That can leave the editor selected on a component that no longer exists.

Could we only call `setSelected(record.id)` when `record.state !== null`?

frontend/src/features/authoring/SceneNavigator/ContextableThumb.tsx
```diff
@@ -80,16 +72,6 @@ function ContextableThumb({
       .catch(handleGeneric);
   };
 
-  async function switchScene(s: Scene) {
```
rgin216, line 83: The old local `switchScene` returned early when `active` was true, but the new handler always calls `switchScene(getScene(), scene._id)`. Clicking the 'already active' scene can now trigger the save/switch path unnecessarily.

I think we should keep the active scene guard either here or inside the shared `switchScene`. But up to u

frontend/src/context/SceneContextProvider.jsx
```diff
@@ -37,13 +38,70 @@ async function saveScenePatch(user, scenarioId, patch) {
...
+        fields[field] = structuredClone(modified[field]);
+      }
+    }
+  );
+
+  return {
+    _id: modified._id,
+    fields,
+    components,
+    deletedComponentIds,
```
rgin216, line 74: `SceneContextProvider` no longer provides `saveScenePatch`, but `EditStateVariable.jsx` still reads and calls it when cleaning deleted state variables out of scene components. This file is not directly changed in the PR, but the provider contract changed here, so the existing cleanup path may silently stop running.

Could we either keep a compatibility `saveScenePatch` in the context or update the state-variable cleanup to use the new `modifyScene` API

frontend/src/features/authoring/AuthoringToolPage.jsx
```diff
@@ -24,28 +26,48 @@ const listeners = [
...
+      modifyScene(getScene());
+      setSaving(false);
+    }, 5000);
+
+    const listener = async ({ operation, record }) => {
+      if (operation === "undo" || operation === "redo") {
+        if (record.sceneId !== sceneId) switchScene(getScene(), record.sceneId);
+        const state = operation === "undo" ? record.before : record.after;
+        if (state === null) setSelected(null);
+        replaceComponent(record.id, state);
```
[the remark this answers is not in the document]
harbassan: switchScene is not async, so the mutation should happen by the time we apply the operations.

frontend/src/features/authoring/AuthoringToolPage.jsx
```diff
@@ -24,28 +26,51 @@ const listeners = [
...
 
-  // autosave setup
+  // NOTE: this is both the autosaver and the history actioner, which are distinct
+  // operations, but are both here due to the limitations of the scene context
   useEffect(() => {
-    if (!sceneId) return;
-    const autosave = setInterval(save, AUTOSAVE_INTERVAL);
-    return () => clearInterval(autosave);
-  }, [sceneId]);
+    const debounced = debounce(async () => {
```
rgin216, line 51: This new autosave path only runs after `historyEvents` fire, but scene-level settings like roles, timer duration, and direct links only call `modifySceneProp` and do not emit history events. Since the old 30-second autosave interval was removed, those edits are no longer saved unless the user manually saves or switches scenes. Navigating away via Back/Resources/Groups/Play after changing these fields can lose data.

frontend/src/features/authoring/scene/history.ts
```diff
@@ -1,87 +1,88 @@
...
+function cloneHistoryRecord(record: HistoryObject): HistoryObject {
+  return {
+    sceneId: record.sceneId,
+    id: record.id,
+    before: structuredClone(record.before),
+    after: structuredClone(record.after),
+  };
+}
+
+const undoStack: HistoryObject[] = [];
```
rgin216: The undo/redo stacks are now module globals that are never reset when the scenario changes. The previous `init` path cleared history when `_scenarioId !== scenarioId`, but that protection was removed. If a user edits scenario A, opens scenario B, then presses undo, the stale record can be applied to the current scene in scenario B after `switchScene` fails to find the old scene id.

frontend/src/context/SceneContextProvider.jsx
```diff
@@ -37,13 +38,70 @@ async function saveScenePatch(user, scenarioId, patch) {
   });
 }
 
+function generatePatch(modified, saved) {
+  const components = [];
+  const deletedComponentIds = [];
+  const fields = {};
+
+  const currentComponents = modified.components ?? {};
```
rgin216, line 46: `generatePatch` assumes `modified.components` is keyed by component id, but some callers pass scenes from React Query where `components` is still an array. For array-shaped components, `Object.entries` uses numeric indexes, `currentComponents[c.id]` is undefined, and the patch can mark every component as both changed and deleted. That turns a targeted update into a full delete/reinsert cycle on the backend.

frontend/src/context/SceneContextProvider.jsx
```diff
@@ -95,25 +153,65 @@ export default function SceneContextProvider({ children }) {
...
 
-  useEffect(() => {
-    if (scenesQuery.data && scenarioId) {
-      init(scenesQuery.data, scenarioId, saveScenePatchWrapper);
-    }
-  }, [scenesQuery.data, scenarioId, saveScenePatchWrapper]);
+  const switchScene = useCallback(
+    (scene, id) => {
+      if (scene._id === id) return; // switch to same scene is a no-op
+      modifyMutationWrapper(scene);
```
rgin216, line 206: Another fire-and-forget. If `parseMedia` or the PATCH request fails while switching scenes, the editor has already replaced the current scene and navigated away, so the user loses the dirty source scene context. We should either await the outgoing save before switching or add an explicit pending-save/recovery path for failed background saves.

rgin216: This is less major as its more of a product/UX tradeoff

rgin216 (dismissed):

frontend/src/context/SceneContextProvider.jsx
harbassan: Ive noted the lack of a recovery system down in the code, but i'm considering this as a future ticket. I feel that the optimistic UI is essential for the usability of the editor.

frontend/src/features/authoring/scene/history.ts
```diff
@@ -1,87 +1,104 @@
...
-  scenes = _scenes;
-  scenarioId = _scenarioId;
-  saveScene = _saveScene;
+
+export const historyEvents = new TypedEventTarget<HistoryEventMap>();
+
+export function clearHistory() {
+  undoStack = [];
+  redoStack = [];
+}
```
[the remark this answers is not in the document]
harbassan: there is no mechanism to move from scenario A to scenario B without unmounting the AuthoringToolPage in the application

leowla (commented):
The save button still exists in the UI, and clicking it shows "Saving". Is the intention to also keep manual saving just in case auto-save fails?

Changes are also lost if the user exits quickly during auto-save.

https://github.com/user-attachments/assets/272787d7-ce36-4979-849a-cfc35f55f16e

harbassan:
> The save button still exists in the UI, and clicking it shows "Saving". Is the intention to also keep manual saving just in case auto-save fails?

Yep, realistically a user would never have to manual save but ive left it there for now just in case we might ever need it.

> Changes are also lost if the user exits quickly during auto-save.

Aware of this, not something i addressed in this PR as its a seperate concern.

rgin216 (approved):
Lgtm

leowla (approved):
💾

PR #431, opened 2026-06-26
fix(autosave): increment pending save inside debounced save

harbassan:
## Issue

The saving status is stuck on "Saving" even after saving is finished.

## Solution

The pending save was being incremented outside the debounce, which means it would be much larger than the actual no. of pending save mutations, so it would never become 0 after successful saves.

## Risk

None

## Checklist

- [ ] Acceptance criteria met

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the saving status indicator so it appears immediately when a save starts.
  * Made the saving state update more reliably during undo and redo actions.

### DIFF
--- a/frontend/src/features/authoring/AuthoringToolPage.jsx
+++ b/frontend/src/features/authoring/AuthoringToolPage.jsx
@@ -50,6 +50,7 @@ export default function AuthoringToolPage() {
   useEffect(() => {
     const debounced = debounce(async () => {
       try {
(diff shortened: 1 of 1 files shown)


harbassan:
@coderabbitai full review

leowla (approved):

leowla (approved):
Is this just the same change?

harbassan:
> Is this just the same change?

yeah just rebase

PR #438, opened 2026-07-05
feat: element hover indicators

harbassan:
## Issue

Hovering over scene elements doesn't give any feedback, which makes it difficult to know what element you're clicking and if youre clicking into text mode or move mode on a textbox. 

## Solution

The cursor now reflects the mode that will be entered upon clicking the component, which is a text cursor for text mode and a move cursor for move mode. All the elements are move mode except the text section of a text box. The hovered state (which comopnent is hovered, if any) is now stored in the editor zustand store, allowing us to render an outline for that component in the overlay. This makes it easy to distinguish which element youre about to click, which is unclear when you have multiple elements stacked.

## Risk

The introduction of the new state value shouldn't cause issues since its only used in one place.

## Checklist

- [ ] Acceptance criteria met

## Summary by CodeRabbit

* **New Features**
  * Hovering over items in the canvas now shows a subtle outline around the hovered item, not just the selected one.
* **UX Improvements**
  * Canvas elements now display more appropriate cursors: move for shapes/images and text cursor for text.
  * Added consistent styling hooks (CSS classes) across box, ellipse, image, speech, and text elements to improve targeting and visual behavior.
* **Bug Fixes**
  * Hover state now updates correctly when moving the mouse (while not pressed).

### DIFF
--- a/frontend/src/features/authoring/canvas/Overlay.tsx
+++ b/frontend/src/features/authoring/canvas/Overlay.tsx
@@ -25,12 +25,15 @@ function resolve(type: Component["type"], bounds: Bounds) {
 
 function Overlay() {
   const selected = useEditorStore((state) => state.selected)!;
+  const hovered = useEditorStore((state) => state.hovered)!;
   const bounds = useEditorStore((state) => state.mutationBounds);
   const scene = useVisualScene((scene) => scene.components);
   const mode = useEditorStore((scene) => scene.mode);
   const createType = useEditorStore((scene) => scene.createType);
 
   const component = scene[selected];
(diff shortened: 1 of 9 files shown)


leowla (approved):
Small visual bug when you hover over an element and leave the canvas the outline will stay.

https://github.com/user-attachments/assets/12315a87-0aff-4573-9fed-4c166bcab21a

And the grey outline overlays on top of the blue outline when the element is selected but you can barely see it so its not useful.

harbassan:
> Small visual bug when you hover over an element and leave the canvas the outline will stay.

Will make a ticket for this

PR #442, opened 2026-07-09
fix: newline  + deleting selected component crash

del-ereno:
## Issue

Deleting a selected textbox with a newline (any empty "" span) inside causes the scenario editor to crash.

## Solution

Fixed element properties guard by adding && component.

## Risk

Hopefully none.

## Checklist

- [x] Acceptance criteria met

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented a crash when rendering text blocks that produce no visual lines.
  * Improved editor behavior when the active scene disappears by clearing the editor before navigating to a valid scene.
  * Kept selection consistent after delete, cut, and Backspace actions by avoiding unnecessary selection clearing when components are removed.
  * Updated undo/redo selection behavior to only change selection when a valid scene state is available.

### DIFF
--- a/frontend/src/features/authoring/AuthoringToolPage.jsx
+++ b/frontend/src/features/authoring/AuthoringToolPage.jsx
@@ -88,7 +88,10 @@ export default function AuthoringToolPage() {
     if (!sceneId || !scenes) return;
     if (!scenes.find((s) => s._id === sceneId)) {
       const next = scenes[0];
-      if (next) replace(next);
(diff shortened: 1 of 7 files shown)


harbassan (approved):
just removed some redundant deselects from before. looks like the issue is actually that the history event is firing a rerender which happens in a cycle prior to the setSelected null cycle. So it can be fixed by just deselecting first, like youve done here.

PR #450, opened 2026-07-25
fix: theme direct-link scene selector

rgin216:
## Issue

The direct-link scene selector used the browser’s native select styling, making it inconsistent with the themed selectors used throughout the authoring interface.

## Solution

Replaced the native selector with the shared themed `SelectInput` component and added disabled-state support to preserve the existing direct-link behavior.

## Risk

Low risk. The shared selector now supports a disabled state, which could affect other usages if they adopt this option incorrectly.

## Checklist

- [x] Acceptance criteria met

## Summary by CodeRabbit

* **New Features**
  * Updated Direct Link settings with a consistent scene-selection dropdown.
  * Added an option to remove a Direct Link target.
  * Added the ability to disable selection controls with improved visual feedback and non-interactive behavior.
  * Enhanced the dropdown trigger with a clearer chevron indicator, improved alignment, and truncated value display.
  * Added a fallback label for scenes without a recognized name.

### DIFF
--- a/frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
+++ b/frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
@@ -1,5 +1,5 @@
-import { useContext, useState, useEffect } from "react";
-import { Check } from "lucide-react";
+import { useContext, useState, useEffect, useRef } from "react";
+import { Check, ChevronDown } from "lucide-react";
 import ScenarioContext from "context/ScenarioContext";
 import SceneContext from "context/SceneContext";
 import { generateUniqueSceneName } from "../../../utils/sceneUtils";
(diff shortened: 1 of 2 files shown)


rgin216:
@coderabbitai

harbassan (changes requested):
sweet, looks good and the accessibility stuff is great. since you added keyboard support though, none of them seem to actually work for me?

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
```diff
@@ -97,6 +106,77 @@ export default function SceneSettings() {
...
+        closeRoles();
+        return;
+      default:
+        return;
+    }
+
+    event.preventDefault();
+    setActiveRoleIndex(nextIndex);
+  }
+
```
harbassan: if we're going to add this for the role multiselect as well, can we move this to a new MultiSelect component ? It gets messy putting all this logic in this file.

harbassan (changes requested):
The up down arrow for moving selected item works when you use enter to focus in on the select, but not when you mouse click to enter into select. Since its an accessibility thing idm that much, but would be good to fix.

harbassan (approved):
yep lgtm after that coderabbit change. 👍 🤠 👍

PR #451, opened 2026-07-26
fix(authoring): context menu flickers at previous position

K1mmyn:
## Issue

~~The main cause for the flicker is that, the current function which create the context menu check onmouseup if your cursor is on a piece of html which is apart of the context menu. This bug happens when you are holding right click and move your cursor off the context menu and release right click.~~

`adjustedPosition` is never cleared when the menu closes so the next open paints one frame at the previous menu's coordinates before the effect corrects it

## Solution

~~Change from mouseup to mousedown because it is guaranteed that the mouse will start on top of the context menu due to negative offset~~

`useLayoutEffect` to adjust first before paint

## Risk

NONE

## Checklist

- [x] Acceptance criteria met

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved context menu click-away handling by responding when the mouse button is pressed, providing more consistent dismissal behavior.

### DIFF
--- a/frontend/src/components/ContextMenu/portal.jsx
+++ b/frontend/src/components/ContextMenu/portal.jsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useRef } from "react";
+import { useState, useEffect, useLayoutEffect, useRef } from "react";
 
 const OFFSET = -5;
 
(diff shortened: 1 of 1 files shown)


GogaIV (approved):
I mean i guesss

harbassan (changes requested):
https://github.com/user-attachments/assets/ebb3e116-cf18-424b-a99f-9c039e301f94

GogaIV (approved):
[Screencast_20260727_235200.webm](https://github.com/user-attachments/assets/26f6cf23-8860-4adb-8eb7-4f2930489b6a)
it is working for me idk hartej, im on firefox browser 
the other sort of flicker thats happening seems to be from the previous context menu location so could not be closing properly or smthn like that

harbassan:
yeah i think that tiny flicker you have at the start of your video is the one im having all the time, where there is a render that happens at the previous location before a rerender at the new location.

harbassan (approved):
works

PR #452, opened 2026-07-27
VPS-120 and VPS-104/set roles from editor

kmck133:
## Issue

Roles could previously only be created as a side effect of uploading a group-assignment CSV. `POST /api/group/:scenarioId` derived the role list from the CSV and overwrote `Scenario.roleList` entirely. This mixed up two separate concerns: defining what roles exist, and assigning people to them. Authors had no way to create roles directly, and scene-level role assignment had nothing to offer until a CSV was uploaded at least once.

Separately, role names only worked reliably when lowercase (VPS-104). Role vocabulary (`Scenario.roleList`, `Scene.roles`) was forced to lowercase on creation, but `Group.users[].role` kept whatever casing appeared in the uploaded CSV. A capitalized role like "Doctor" in a CSV would silently fail scene-access checks against the lowercased `Scene.roles`.

## Solution

Made roles a first-class, editor-managed list, decoupled from CSV upload, and fixed the case-sensitivity bug:

- Added `GET/POST/DELETE /api/scenario/:scenarioId/roles`, mirroring the existing `stateVariables` CRUD pattern. Creation trims and dedupes case-insensitively (existing casing wins on conflict). Deletion cascades a `$pull` across the scenario's scenes so deleted roles don't linger as orphaned restrictions.
- CSV upload now merges new roles into the existing `Scenario.roleList` instead of overwriting it, so editor-created roles survive a later CSV upload.
- Removed the forced lowercasing of role names at every creation point (editor and CSV upload). Roles now display with whatever casing they were authored with.
- Made scene-access role checks (`getSceneConsideringRole`, `getConnectedScenes` in `navigate/group.js`) case-insensitive and whitespace-insensitive via a shared `roleMatches` helper, so mismatched casing or a stray space from an uploaded CSV no longer locks players out of a scene.
- Added a "Roles" modal (`RoleMenu` / `CreateRole`), accessible from the Manage Groups page, with a compact table listing roles and per-row delete. Supports Enter-to-create and delete-without-closing-the-dialog, and blocks duplicate names (case-insensitively) client-side with a clear error toast.
- `ScenarioContext` now sources `roleList` from the new scenario-owned endpoint instead of the group/CSV-coupled one.

Fixes VPS-104.

## Risk

- No rename support. Role names are duplicated as plain strings across `Scene.roles`, `Scenario.roleList`, `Group.users[].role`, and `Note.role` with no shared ID. Renaming safely would need a cascade that only `Scene.roles` can support. For now, renaming means delete, recreate, and reassign.
- Deleting an in-use role cleans up `Scene.roles`, but not `Group.users[].role` or `Note.role`. Those go stale if a role is deleted after groups have already been assigned via CSV.
- CSV upload used to overwrite `Scenario.roleList` on every upload. Now it merges. Unlikely to affect real workflows but worth flagging.
- Old `GET /api/group/:scenarioId/roleList` is kept for backward compatibility instead of removed, so no existing consumers break.

## Checklist

- [ ] Acceptance criteria met

## Summary by CodeRabbit

* **New Features**
  * Added scenario role management (view, create, delete) with a new Roles menu in the Manage Groups page.
* **Improvements**
  * Role names are merged case-insensitively while preserving original capitalization.
  * Role inputs accept surrounding whitespace; CSV role imports trim, dedupe case-insensitively, and ignore empty entries.
  * Role-based navigation access now ignores casing and stray leading/trailing whitespace.
* **Bug Fixes**
  * Prevented duplicate roles when adding via uploads or role creation.
  * Deleting a role now removes it from related scenes.
* **Tests**
  * Expanded API and navigation authorization test coverage for role merging and casing/whitespace behavior.


harbassan (dismissed):
sweet, code looks clean and works well. I made a little change to use the existing normalisation function for normalising.

the only thing i noticed is the input element doesn’t expand to the width of the modal:

Image

harbassan (approved):
---😁  
👍 👕 👍
--🦶 🦶

PR #462, opened 2026-08-03
fix: make resource collection popup work in Safari

glybrio:
## Issue

The resource collection creation popup did not appear when clicking Create in Safari, although it worked correctly in other browsers.

The interface also used inconsistent terminology, referring to the same feature as both Collections and Groups.

## Solution

Updated the dropdown so its open state is controlled explicitly using React state and DaisyUI’s dropdown-open class, rather than relying on browser focus behaviour. This ensures the popup displays consistently in Safari and other supported browsers.

Also updated the popup wording to consistently use Collection instead of Group.

## Risk

Low risk apart from Hartej's colouring books.

## Summary by CodeRabbit

* **New Features**
  * Automatically focuses the collection name field when opened.
  * Added Enter and Escape keyboard controls.
  * Supports closing when clicking outside the dialog.
  * Improved dropdown accessibility with appropriate attributes.
* **Bug Fixes**
  * Collection names are trimmed and validated consistently.
  * Empty-name submissions are disabled.
  * Form state resets when the dialog closes.
* **Style**
  * Updated labels and button text to use “collection” terminology.

### DIFF
--- a/frontend/src/features/resources/components/AddGroup.jsx
+++ b/frontend/src/features/resources/components/AddGroup.jsx
@@ -40,11 +40,7 @@ function AddGroup({ onAdd }) {
   };
 
   return (
-    <div
-      className={`dropdown dropdown-end ${
-        open ? "dropdown-open" : ""
-      }`}
-    >
+    <div className={`dropdown dropdown-end ${open ? "dropdown-open" : ""}`}>
       <button
         type="button"
         className="btn btn-phantom btn-sm"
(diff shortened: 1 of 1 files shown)


leowla (commented):
~~Since the dropdown is a DaisyUI component I wonder if this would be fixed by just updating the dependency since we are on a beta version IIRC so maybe we won't need a useEffect to workaround this.~~
Edit: I am wrong

leowla (approved):
Fixed the Safari issue, but clicking outside the popup no longer closes it.

leowla (approved):
🧭

PR #464, opened 2026-08-03
fix(authoring): spawn new elements at mouse position

glybrio:
## Issue

When a new canvas element was created by clicking instead of dragging, it appeared in the top left corner of the canvas rather than at the user's mouse position.

## Solution

Updated `createComponentFromBounds` so that elements created with a click or small drag keep their default dimensions but are translated to the initial mouse position.

Elements created by dragging more than the minimum size continue to use the dimensions defined by the drag action.

## Risk

Low risk. But harbassan is the big risk.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected component placement when creating components with small dimensions.
  * Components now appear at the requested origin while retaining appropriate default sizing.

### DIFF
--- a/frontend/src/features/authoring/scene/operations/component.ts
+++ b/frontend/src/features/authoring/scene/operations/component.ts
@@ -132,12 +132,19 @@ export function createComponentFromBounds(
   bounds: Bounds
 ) {
   const component = structuredClone(defaults[type]);
+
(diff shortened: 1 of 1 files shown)


harbassan (approved):
luvly 😍 ✊

PR #469, opened 2026-08-05
Vps 61 infinite toast fix

FridgeProtector1:
## Issue

The infinite toast bug occurs because of a recursive loop in the 409 error handling

## Solution

ref-based guard to prevent recursive 409 error handling

## Risk

Nada.

## Checklist

- [x] Acceptance criteria met

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multiplayer and single-player navigation after conflicts by refreshing scene and game state data.
  * Prevented duplicate conflict-resolution requests and outdated navigation responses during rapid actions.
  * Ensured the correct scene and remaining time are displayed after navigation conflicts or refreshes.
  * Prevented automatic scene changes when manual conflict resolution is required.
  * Added fallback routing to the error page if synchronization fails.
  * Preserved scene timer progress when returning to cached scenes.
  * Improved resource display using the updated overlay.


frontend/src/features/playScenario/PlayScenarioPage.jsx
```diff
@@ -132,12 +133,15 @@ export default function PlayScenarioPage({ group }) {
...
-      );
+      if (!handlingConflictRef.current) {
+        handlingConflictRef.current = true;
+        toast.success(
+          isMultiplayer
+            ? "Someone else made a move first, but you're back on track!"
+            : "A move from somewhere else was made, but you're back on track!"
+        );
+        setTimeout(() => { handlingConflictRef.current = false; }, 1000);
+      }
```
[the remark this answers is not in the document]
FridgeProtector1: I dont get it, what is the intended behavior of the game? Someone smart save me

harbassan (changes requested):
yeah u need to actually make the request to get the new state, which is what it was doing before. The way youve done it it just doesn't make a refresh request, so nothing happens. 

Basically: on 409, get the actual current state, then toast after switch.

You have: on 409, toast.

frontend/src/features/playScenario/PlayScenarioPage.jsx
```diff
@@ -84,6 +84,51 @@ const navigateMultiplayer = async (
...
+  const res = await axios.request(config);
+  if (res.data.scenes) {
+    res.data.scenes.forEach((scene) => sceneCache.set(scene._id, scene));
+  }
+  return {
+    newSceneId: res.data.active,
+    stateVariables: res.data.stateVariables,
+    newStateVersion: res.data.stateVersion,
+  };
+};
```
harbassan, line 134: i may be missing something, whats the point of this function compared to `onSceneChange`? Looks like it does the same thing?

harbassan (changes requested):

frontend/src/features/playScenario/PlayScenarioPage.jsx
FridgeProtector1: onSceneChange sends active move data like componentId and flags to progress the player forward. refreshFromServer is just for 409 error recovery it sends null for the scene so we can fetch the server's true state without accidentally making a new move or applying state changes. But like onScene change could be refactored to handle both if u want.

FridgeProtector1: Cause the infinite toast was caused by onSceneChange being called with the same parameters recursively.

harbassan (approved):
lgtm

harbassan (approved):
works

PR #470, opened 2026-08-08
VPS-162 and VPS-163/Pasting text in textbox

kmck133:
## Issue

The textbox authoring tool had several text editing bugs.

- Pasted text kept the font and style it had when copied instead of matching the text it was pasted into. (Addresses VPS-163)
- Undo or redo shortly after an edit could crash the app with "Cannot read properties of undefined (reading 'style')".
- Undo and redo did not put the cursor back where the change happened, so continuing to type required clicking back into place manually.
- Pasting while a range of text was selected did not replace the selection. It inserted the new content and left the old selected text in place. (Addresses VPS-162)
- Double clicking a word, or clicking a third time to select everything, did nothing.

## Solution

- Paste now looks up the style of the text surrounding the insertion point and applies that to the pasted content, instead of keeping the style from where it was copied.
- The undo and redo handler now clears the text selection before the document is replaced instead of after. React does not batch these state updates outside of event handlers, so a render could happen with a selection that no longer matched the restored document.
- Undo and redo now compare the before and after versions of the document to find exactly which characters changed, then place the cursor or selection there. This matches how undo works in other text editors, including selecting text that reappears and collapsing the cursor where text disappears.
- Paste now deletes the active selection before inserting, the same way typing over a selection already worked.
- Mouse clicks on text now check the click count. A double click selects the word under the cursor, and a triple click selects the whole document, reusing the existing select all logic.

## Risk

- Paste always overwrites the style of pasted content with the destination style. Pasting a selection that intentionally had mixed formatting inside it, such as one bold word in an otherwise plain sentence, will lose that internal formatting and become uniform.
- Undo and redo now force the affected textbox into text edit mode and move the selection to the edit location, even if the user was doing something else, such as resizing a different shape, at the time.
- Word selection uses a Unicode aware character set for letters and numbers, but the definition of a word boundary is still a simplification and may not match every language's expectations.
- There is no automated test suite for this module, so these changes were verified with linting, type checking, a production build, and manual testing in the browser rather than unit or integration tests.

## Checklist

- [ ] Acceptance criteria met

## Summary by CodeRabbit

- **New Features**
  - Added double-click word selection and triple-click select-all in the text editor.
  - Improved cursor and selection synchronization when switching between text and visual editing modes.
  - Enhanced cursor movement across lines and styled text boundaries.

- **Bug Fixes**
  - Undo and redo now restore text selections more accurately after scene changes.
  - Pasting validates content, safely falls back to plain text, and correctly replaces active selections.
  - Pasted text now adopts the formatting at the insertion point instead of retaining unintended source formatting.


frontend/src/features/authoring/scene/operations/text.ts
```diff
@@ -198,6 +198,176 @@ function normaliseSelection(sel: ModelSelection) {
...
+        inBlock -= len;
+      }
+      return { blockI, spanI: 0, charI: 0 };
+    }
+
+    remaining -= text.length + 1; // +1 for the "\n" block break
+  }
+
+  return { blockI: 0, spanI: 0, charI: 0 };
+}
```
harbassan, line 241: instead of doing the cursor to offset conversion and then using the offset and converting back to a cursor, can we just use the cursor directly? The parseHit already gives the cursor position of the click, so it would be simpler to just walk back and forward through the visual document by spanI/charI . We have move cursor visual function you can use to walk left and right.

harbassan (changes requested):
everything looks good except the full word select on double click functionality.

frontend/src/features/authoring/scene/operations/text.ts
kmck133: used moveCursorVisual and also fixed a small bug with it

frontend/src/features/authoring/handlers/pointer/pointer.ts
```diff
@@ -180,29 +177,105 @@ function handleDocumentClick(e: React.MouseEvent, position: Vec2) {
...
-  const text = flattenBlocks(blocks);
-  const offset = cursorToOffset(blocks, selection.start);
+function sameVisualCursor(a: VisualCursor, b: VisualCursor) {
+  return (
+    a.blockI === b.blockI &&
+    a.lineI === b.lineI &&
+    a.spanI === b.spanI &&
+    a.charI === b.charI
+  );
+}
```
harbassan: we dont need this function, you can just perform an equality using fastIsEqual package we use

frontend/src/features/authoring/handlers/pointer/pointer.ts
```diff
@@ -180,29 +177,105 @@ function handleDocumentClick(e: React.MouseEvent, position: Vec2) {
...
 
-  const { start, end } = findWordRange(text, offset);
-  if (end <= start) return;
+// walks in one direction while the adjacent char is a word char. the
+// sameVisualCursor check is just a safety net -- moveCursorVisual should
+// always make progress here now, but a walk that trusted that blindly is
+// exactly what caused the freeze this was built to fix in the first place
```
harbassan: seems like this ai comment is useless out of context

harbassan (changes requested):
just a couple things, but functionality works great

harbassan (approved):
lgtm

PR #471, opened 2026-08-08
fix(authoring): proportional resize axis lock

leowla:
## Issue

Corner resize with shift/ctrl compared raw dx/dy against the fixed opposite corner instead of each axis's own dimension, so the larger original axis (e.g. width on a wide shape) almost always "won", leaving the shorter axis effectively locked during drag.

## Solution

Compare movement relative to each axis's own original size

## Risk

Low

## Checklist

- [x] Acceptance criteria met

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved aspect-ratio locking during resizing, producing more consistent results when dragging horizontally or vertically.


harbassan (approved):
lgtm

PR #475, opened 2026-08-10
fix: lock image upload to originating slide

glybrio:
## Issue

Image uploads were not locked to the slide where the upload started.

If a user started uploading an image on one slide and navigated to another slide before the upload completed, the image could be added to the newly selected slide instead of the original slide.

## Solution

Capture the originating scene before the asynchronous image upload or image loading begins.

When the image is ready:
- If the user is still on the originating slide, add the image normally.
- If the user has moved to another slide, save the image to the originating scene instead.

This also applies the same behaviour when selecting an existing image.

## Risk

Low risk. (Apart from Hartej vs Leo 
image
)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image insertion to ensure images are added to the scene where they originated, even if you navigate to another scene while they load.
  * Applied consistent behavior to both uploaded and existing images.
  * Image additions now retain their source context and unique identity for more reliable editing.

### DIFF
--- a/frontend/src/features/authoring/images.tsx
+++ b/frontend/src/features/authoring/images.tsx
@@ -6,23 +6,54 @@ import { useParams } from "react-router-dom";
 import { ImageIcon } from "lucide-react";
 import { add } from "./scene/operations/modifiers";
 import { defaults } from "./scene/operations/component";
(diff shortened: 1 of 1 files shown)


harbassan (approved):
Image

PR #478, opened 2026-08-11
Vps 161/shapes cant have text directly within them

GogaIV:
## Issue

All shapes need to be able to create a Text field inside them on double click (i initially thought it was only box's only lol)

## Solution

textbox has a Text field which also requires a document object so shapeComponent now has a document object but unlike textbox it is optionally so it is only created when the user double clicks on the object, itll also work on old shapes too as it creates the document prior to mapping to canvas

## Risk

nothing from here has caused errors on my end, there is an inherent problem with textbox's being able to be letter spammed beyond their width, and furthermore there is the fact that on ellipses you can go outside the circular shape but i believe the user should be allowed freedom to extend the text beyond the limits of the shape if they wish so im not changing that, though it is noted

## Checklist

- [ ] Acceptance criteria met

## Summary by CodeRabbit

* **New Features**
  * Added text support to box, ellipse, and speech-bubble shapes.
  * Users can add and edit text directly within these shapes by double-clicking.
  * Text styling controls are available for shapes containing text.
  * New text-enabled shapes display placeholder content when text editing begins.

* **Improvements**
  * Text content and formatting are preserved during rendering.
  * Shapes without text content continue to work as before.
  * Selected text components now retain their active styling.

* **Bug Fixes**
  * Corrected padding behavior for zero-valued directions.

### DIFF
--- a/frontend/src/features/authoring/components/AddText.tsx
+++ b/frontend/src/features/authoring/components/AddText.tsx
@@ -1,0 +1,25 @@
+import Text from "../text/Text";
+import { modifyComponentProp } from "../scene/operations/component";
+
+export function addText(WrappedComponent) {
+  return function TextableShape(props) {
+    if (!props.document) {
+      modifyComponentProp(props.id, "document", {
+        style: {},
+        blocks: [{ style: {}, spans: [{ style: {}, text: "" }] }]
+      });
+    }
+    function handleDoubleClick(e: React.MouseEvent) {
+      const rect = e.currentTarget.querySelector('[data-type="document"]') as HTMLElement;
+      if (!rect) return;
+      rect.dispatchEvent(new MouseEvent('mousedown', { bubbles: true, clientX: e.clientX, clientY: e.clientY }));
+    }
+
+    return (
+      <g onDoubleClick={handleDoubleClick}>
+        <WrappedComponent {...props} />
+        {props.document && <Text doc={props.document} editable={true} />}
+      </g>
+    );
+  }
+}
--- a/frontend/src/features/authoring/components/AddText.tsx
+++ b/frontend/src/features/authoring/components/AddText.tsx
@@ -1,18 +1,27 @@
 import Text from "../text/Text";
 import { modifyComponentProp } from "../scene/operations/component";
+import type { ShapeComponent } from "../types";
 
-export function addText(WrappedComponent) {
-  return function TextableShape(props) {
+export function addText(WrappedComponent: ShapeComponent) {
+  return function TextableShape(props: ShapeComponent) {
     if (!props.document) {
       modifyComponentProp(props.id, "document", {
         style: {},
-        blocks: [{ style: {}, spans: [{ style: {}, text: "" }] }]
+        blocks: [{ style: {}, spans: [{ style: {}, text: "" }] }],
       });
     }
     function handleDoubleClick(e: React.MouseEvent) {
-      const rect = e.currentTarget.querySelector('[data-type="document"]') as HTMLElement;
+      const rect = e.currentTarget.querySelector(
+        '[data-type="document"]'
+      ) as HTMLElement;
       if (!rect) return;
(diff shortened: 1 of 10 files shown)


harbassan (changes requested):
~~I am a disliker of this approach ☹️. It doesn't allow us to extend the system further than the simplest case that's implemented here, iykwim.~~ 

Revised: can we extend this further so that text within all elements is treated equally to text within a textbox?

harbassan (approved):
lgtm

harbassan:
@coderabbitai review

GogaIV:
i forgot to add bounds for ellipse it is NOT done!!!

leowla (changes requested):
1. Toolbar should show text style options when a shape with text is selected
2. Long words should wrap to shape
3. Shape text should default to being vertically and horizontally centred inside the shape.

Actual
Image

Expected
Image

frontend/src/features/authoring/components/AddText.tsx
```diff
@@ -0,0 +1,34 @@
+import Text from "../text/Text";
+import { modifyComponentProp } from "../scene/operations/component";
+import type { ShapeComponent } from "../types";
+
+export function addText(WrappedComponent: ShapeComponent) {
+  return function TextableShape(props: ShapeComponent) {
+    if (!props.document) {
+      modifyComponentProp(props.id, "document", {
+        style: {},
+        blocks: [{ style: {}, spans: [{ style: {}, text: "" }] }],
```
K1mmyn, line 10: You should add a placeholder when you first create the textbox because it hard to tell that the textbox has been created

frontend/src/features/authoring/topbar/Topbar.tsx
```diff
@@ -9,69 +9,60 @@ import ShapeSection from "./ShapeSection";
...
 import ShapeCreateMenu from "./ShapeCreateMenu";
-import BackgroundMenu from "./BackgroundMenu";
-
+import type { Component } from "../types";
 import "./topbar.css";
 
+function hasDocument(component: Component): boolean {
+  return (
+    "document" in component &&
+    !!component.document &&
```
K1mmyn, line 24: remove the !!

frontend/src/features/authoring/topbar/Topbar.tsx
```diff
@@ -9,69 +9,60 @@ import ShapeSection from "./ShapeSection";
...
+    component?.document.blocks?.[0]?.spans?.[0]?.text?.length > 0
```
K1mmyn, line 25: make this variable instead of long chain to improve readability

frontend/src/features/authoring/pipeline.ts
```diff
@@ -26,16 +26,25 @@ function pad(verts: Vec2[], amount: number) {
   const center = getBoxCenter(verts);
   return verts.map((vert) => {
     const relative = subtract(center, vert);
-    const dir = mutate(relative, (val) => val / Math.abs(val));
+    const dir = mutate(relative, (val) =>
+      val === 0 ? 0 : val / Math.abs(val)
```
K1mmyn, line 30: W division by 0 catch by code rabbit himself

frontend/src/features/authoring/scene/operations/component.ts
```diff
@@ -65,6 +60,20 @@ export const defaults = {
...
+          style: {},
+          spans: [
+            {
+              style: {},
+              text: "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla venenatis.",
+            },
+          ],
+        },
+      ],
+    },
```
K1mmyn, line 76: can you add some padding by adding the placeholding in the 5 row or smth

K1mmyn (changes requested):
LGTM but there is a bug!!!

frontend/src/features/authoring/topbar/Topbar.tsx
GogaIV: lint wont pass otherwise as component.document is not a bool by default, I have to cast it to bool, but i used boolean() so it makes more sense

PR #493, opened 2026-08-17
VPS-96/change name of uploaded resources

kmck133:
## Issue

Users had no way to rename an uploaded resource from the Manage Resources page. Fixing a typo or renaming a file required deleting it and re-uploading, losing any state conditionals or references already set up on it. Separately, the "Download" button on a resource's preview didn't reliably download the file, meaning for cross-origin file URLs (e.g. Firebase Storage), the browser would just open/navigate to the file instead of saving it.

## Solution

- Added `PATCH /api/resources/:scenarioId/:resourceId` to rename a resource (file or collection), with server-side validation (non-blank, ≤255 chars).
- File resources have their extension locked. The rename endpoint rejects any attempt to change a file's extension, so it can't accidentally be mislabeled as a different type.
- Added inline rename UI to the resources list: a pencil icon toggles into a text input, with the file extension shown as a fixed, non-editable suffix. A checkmark button replaces the delete button while editing, so an accidental click can't delete the resource mid-rename.
- Fixed text overflow: long resource names now truncate with an ellipsis (and wrap instead of overflowing in the preview panel) rather than pushing the delete/edit controls off-screen.
- Fixed the resource preview's Download button: it now fetches the file and downloads it via a blob URL instead of relying on the native `download` attribute, which browsers ignore for cross-origin file URLs.

## Risk

- File renames are now validated server-side (255-char limit, extension lock) — any existing script or integration that renames resources with a different extension will start receiving a 400 where it previously succeeded.
- Collections (groups) can be renamed via the API and are covered by a backend test, but adding a button for it in the app seemed like a separate issue that was out of scope for this ticket, meaning that only file resources are renameable from the page for now.
- Download now fetches the whole file into memory client-side before saving it, rather than letting the browser stream it directly. This is worth keeping an eye on for very large files.

## Checklist

- [ ] Acceptance criteria met

## Summary by CodeRabbit

* **New Features**
  * Added inline renaming for files and collections from the resource management page.
  * Resource names are trimmed and limited to 255 characters.
  * Added safer resource downloads with file-type labels and appropriate filenames.
* **Bug Fixes**
  * Improved display of long names with wrapping and truncation.
  * Rename failures now restore the previous name and refresh resource data.
  * Added clearer feedback when downloads or rename actions fail.

### DIFF
--- a/backend/src/routes/api/resources.js
+++ b/backend/src/routes/api/resources.js
@@ -17,12 +17,6 @@ const router = Router();
 
 router.use(auth);
 
-function getExtension(name) {
-  const trimmed = name.trim();
-  const dotIndex = trimmed.lastIndexOf(".");
-  return dotIndex <= 0 ? "" : trimmed.slice(dotIndex).toLowerCase();
-}
-
 /**
  * @route GET /api/resources/:scenarioId
  * @desc Get all resources for a scenario
@@ -34,7 +28,7 @@ router.get(
     const { scenarioId } = req.params;
 
     const resources = await Resource.find({ scenarioId })
-      .populate("fileId", "url type contentType size")
+      .populate("fileId", "name url type contentType size")
       .sort({ parentId: 1, createdAt: -1 })
       .lean();
 
@@ -88,7 +82,7 @@ router.post(
       name,
       fileId,
     });
-    await resource.populate("fileId", "url type contentType size");
+    await resource.populate("fileId", "name url type contentType size");
 
     await applyReferenceDelta(fileId, 1);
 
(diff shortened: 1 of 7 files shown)


backend/src/routes/api/resources.js
```diff
@@ -156,6 +164,51 @@ router.delete(
...
+
+    if (
+      existing.type === "file" &&
+      getExtension(trimmedName) !== getExtension(existing.name)
+    )
+      throw new HttpError(
+        "file extension cannot be changed",
+        HttpStatusCode.BadRequest
+      );
+
```
harbassan: we actually are fine with not having a dot extension, so you can turn this into an atomic update. one of the main points of the rename is that the file extension can be removed, so the user sees something like "Patient Info Sheet" instead of "patient_info_sheet.pdf".

frontend/src/features/resources/ManageResourcesPage.jsx
```diff
@@ -345,6 +392,141 @@ export default function ManageResourcesPage() {
 }
 
 // Helper components
+function splitFileName(name) {
+  const dotIndex = name.lastIndexOf(".");
+  if (dotIndex <= 0) return { base: name, ext: "" };
+  return { base: name.slice(0, dotIndex), ext: name.slice(dotIndex) };
+}
+
+function ResourceNameField({
```
harbassan: can we move this component to its own file pls

harbassan (changes requested):
functionally it works well. 👍 besides the comments, can you revert the style changes so that the file tree looks like it did before?

also just for future ref the download behaviour change should be a separate pr.

backend/src/routes/api/resources.js
kmck133: made it so the user can always remove the extension and name it whatever they want, but the original filetype always remains. now if the user downloads the png file `image.png`, it'll be a file called `image` with png type but if they rename it `image.pdf`, it'll download as a file called `image.pdf` with png file type. i also added a little badge in the preview of the file so the user knows what kind of file they're downloading

harbassan (approved):
lgtm

kmck133:
@coderabbitai review

PR #494, opened 2026-08-17
VPS-128/Custom Keyboard Mapping for Navigable Components

kmck133:
## Issue

Buttons in a scenario could only ever be activated by mouse click. There was no way for an author to navigate with their keyboard apart from the "Direct Link" button which had its limitations. The existing "Direct Link" advance-to-next-scene feature was hardcoded to always respond to Space or the right arrow key with no way to customize it.

## Solution

- Authors can now bind any clickable component to a key from a curated, safe set (letters, digits, common punctuation, Space, and the arrow keys, nothing reserved for browser/OS shortcuts), picked via a "press a key to bind" capture control instead of a long dropdown.
- An optional on-screen key hint badge (`[W]`) can be shown next to a bound button, with a configurable position (6 spots around the button), visible both in the authoring canvas and during actual play.
- Direct Link is now unified into the same key-binding system instead of being a separate, hardcoded mechanism: by default it still responds to Space or Right Arrow (identical to today's behavior, so existing scenes need no migration), or an author can switch it to one specific custom key.
- Collision prevention happens at multiple layers: the authoring pickers dynamically exclude keys already claimed elsewhere in the scene; switching Direct Link between Default/Custom mode automatically clears any component binding that would otherwise silently collide; and a new server-side check (`assertUniqueKeyBindings` in `sceneDao.patchScene`) rejects any save that would leave two components, or a component and the direct link, bound to the same key.

## Risk

- `Scene.directLinkKey` is a new, additive schema field (defaults to `null`); no data migration is needed since `null`/unset still means "respond to Space or Right Arrow," exactly matching every existing scene's current behavior.
- The new server-side `assertUniqueKeyBindings` check is a new 400 rejection path, which is a scene patch that would create a key collision now fails at save time instead of silently succeeding. This is intentional (data integrity), but is a new failure mode; worth confirming the frontend surfaces a clear error rather than a generic "something went wrong" toast if it's ever hit in practice.
- No impact to scenes/components not using this feature. All new fields are optional and additive on top of the existing shape.

## Checklist

- [x] Acceptance criteria met

## Summary by CodeRabbit

* **New Features**
  * Assign keyboard shortcuts to clickable components, with optional on-canvas key hints.
  * Configure the key used for direct scene navigation; Space and Arrow Right remain the defaults.
* **Bug Fixes**
  * Prevent conflicting shortcuts and clear outdated navigation or key-hint settings when they’re no longer valid.
  * Preserve valid component bindings when scenes are edited or removed.
  * Show “None” when a linked scene is unavailable.


harbassan:
@leowla im taking time off from reviewing, this ones all yours

backend/src/db/daos/__tests__/sceneDao.test.js
```diff
@@ -3,13 +3,16 @@ import { describe, beforeEach, it, expect } from "@jest/globals";
 import mongoose from "mongoose";
```
K1mmyn, line 4: There is a bug where the key hint doesn't show

https://github.com/user-attachments/assets/85001734-cd12-4ec2-babf-ea7014a443f1

This happens only when you copy paste components

Next Steps: Resolve on your own or make this a new ticket

K1mmyn: Possible concern, the hint shows uppercase letters but bindings can only be lowercase letter

frontend/src/features/authoring/keyBindings.ts
```diff
@@ -0,0 +1,121 @@
+// Single, unmodified keys an author may bind a button (or a scene's direct
+// link) to during playback. Excludes anything still reserved: Enter/Tab/
+// Escape/Backspace/Delete/F-keys/modifiers, and "/" and "'" (both trigger
+// Firefox's Quick Find when pressed unmodified).
+const LETTERS = Array.from({ length: 26 }, (_, i) =>
+  String.fromCharCode(65 + i)
+);
+const DIGITS = Array.from({ length: 10 }, (_, i) => String(i));
+const PUNCTUATION = ["-", "=", "[", "]", ";", ",", ".", "`", "\\"];
+
```
K1mmyn, line 13: You can add " ' " to the list right?

frontend/src/features/authoring/keyBindings.ts
```diff
@@ -0,0 +1,121 @@
...
+  ARROWLEFT: { nativeKey: "ArrowLeft", display: "←" },
+  ARROWRIGHT: { nativeKey: "ArrowRight", display: "→" },
+};
+
+const NATIVE_KEY_TO_ID: Record<string, string> = Object.fromEntries(
+  Object.entries(NAMED_KEYS)
+    .filter(([, { nativeKey }]) => nativeKey)
+    .map(([id, { nativeKey }]) => [nativeKey as string, id])
+);
+
```
K1mmyn, line 31: NAMED_KEYS and NATIVE_KEY_TO_ID seems a little unnecessary. why not just have something like

{
  "ArrowUp": "ARROWUP",
  "ArrowDown": "ARROWDOWN",
  "ArrowLeft": "ARROWLEFT",
  "ArrowRight": "ARROWRIGHT"
}

Unless there is a reason why you have done this that I am unaware of, I would suggest changing it because you get the same product but it makes the code alot more readable and easier to understand.

backend/src/db/models/scene.js
```diff
@@ -38,6 +38,12 @@ const sceneSchema = new Schema({
     ref: "Scene",
     default: null,
   },
+  // The key that triggers directLink; null means the default of Space or
+  // ArrowRight (see frontend/src/features/authoring/keyBindings.ts).
+  directLinkKey: {
+    type: String,
+    default: null,
+  },
```
K1mmyn, line 98: insanely nitpicking comment but having default: null mean default keybind doesn't really make sense

frontend/src/features/authoring/keyBindings.ts
```diff
@@ -0,0 +1,121 @@
...
+  ...PUNCTUATION,
+  ...Object.keys(NAMED_KEYS),
+];
+
+export function displayKeyBinding(key: string): string {
+  return NAMED_KEYS[key]?.display ?? key;
+}
+
+export function normalizeEventKey(e: KeyboardEvent): string | null {
+  if (e.ctrlKey || e.metaKey || e.altKey) return null;
```
K1mmyn, line 44: it could make sense to add shiftkey to this check since we are only accepting lowercase letters

frontend/src/features/authoring/CanvasSideBar/ComponentProperties.jsx
```diff
@@ -1,10 +1,35 @@
...
+const orDefaultPosition = (v) => v ?? DEFAULT_KEY_HINT_POSITION;
+
+// Mirrors a component prop field into local editable state so edits survive
+// re-renders, resyncing only when the field's value actually changes on the
+// incoming component (preserves in-flight local edits otherwise).
+function usePropMirror(component, field, normalize) {
+  const [value, setValue] = useState(() => normalize(component?.[field]));
+  useEffect(() => {
+    const next = normalize(component?.[field]);
+    setValue((prev) => (prev === next ? prev : next));
```
K1mmyn: how come you have used a ternary operator here? Are you trying to prevent a rerender? if not would be best to just do setValue(prev => next) because more readable

backend/src/db/daos/__tests__/sceneDao.test.js
K1mmyn: https://github.com/user-attachments/assets/edd1b785-ca06-4725-8ac0-c567d2d4e37c

When you create these links to a slide and you delete the slides, keybinds dont get unset

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
```diff
@@ -191,28 +245,54 @@ export default function SceneSettings() {
...
-              }
-              display={(targetId) =>
-                scenes?.find((scene) => scene._id === targetId)?.name ??
-                "Unknown scene"
-              }
-              onChange={(targetId) =>
-                modifySceneProp("directLink", targetId || null)
-              }
-            />
+            {directLink && !directLinkDisabled && (
```
K1mmyn, line 256: edge case, this should not be open if there is only one scene

K1mmyn: https://github.com/user-attachments/assets/7ebc6116-7bc0-4eb4-b499-710153e025c8

This will cause the app the crash because when you click space it will navigate to a non-existent scene

backend/src/db/daos/__tests__/sceneDao.test.js
```diff
@@ -163,6 +178,106 @@ describe("Scene DAO patchScene tests", () => {
     expect(updatedScene.name).toBe("Updated Scene Name");
     expect(updatedScene.roles).toEqual(["patient"]);
     expect(updatedScene.time).toBe(120);
+    expect(updatedScene.directLinkKey).toBe("W");
     expect(updatedScene.components).toHaveLength(3);
   });
+
```
K1mmyn, line 185: Lowkey don't know how this work ngl so the next person please review this :)

K1mmyn (commented):
LGTM! But I have found some bug which should be easy to fix hopeful.

backend/src/db/daos/__tests__/sceneDao.test.js
K1mmyn: Can you also check that if there is a button which goes to the same scene if it resets state variable to default values?

frontend/src/features/authoring/keyBindings.ts
K1mmyn: ignore this, idk what im on about 😭

backend/src/db/daos/__tests__/sceneDao.test.js
K1mmyn: Please bring this up in the next meeting but we need to disable the create button after each click because otherwise you can click it 500 bilion times and it creates 500 billion new pages with the same name

frontend/src/features/authoring/keyBindings.ts
```diff
@@ -0,0 +1,121 @@
...
+interface KeyBoundComponent {
+  id: string;
+  clickable?: boolean;
+  keyBinding?: string | null;
+}
+
+interface ActionableComponent {
+  clickable?: boolean;
+  nextScene?: string | null;
+  stateOperations?: unknown;
```
leowla: Be type safe by reusing the type in authoring types.ts

frontend/src/features/authoring/keyBindings.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Whether a component actually does something when clicked/keyed, as opposed
+// to being clickable with nothing wired up yet. Shared so the editor canvas,
+// the play canvas, and the play keydown handler can't drift on what counts
+// as "actionable".
+export function hasClickAction(component: ActionableComponent): boolean {
+  return !!(
+    component.clickable &&
+    (component.nextScene || component.stateOperations)
```
leowla:
```suggestion
    (component.nextScene || (component.stateOperations?.length ?? 0) > 0)
```
an empty array would still render as true

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
```diff
@@ -174,6 +225,9 @@ export default function SceneSettings() {
                     defaultDirectLinkScene ??
                     scenes?.find((s) => s._id !== selfId)?._id ??
                     null;
+
+                  if (!directLinkKey) clearDefaultDirectLinkCollisions();
```
leowla:
```suggestion
                  if (directLinkKey == null) clearDefaultDirectLinkCollisions();
```
falsy check is not the same as null check which you've used everywhere else

backend/src/db/daos/sceneDao.js
```diff
@@ -308,23 +356,71 @@ const updateSceneOrder = async (scenarioId, sceneIds) => {
...
+      : existingScene.directLink;
+  const effectiveDirectLinkKey =
+    "directLinkKey" in allowedFields
+      ? allowedFields.directLinkKey
+      : existingScene.directLinkKey;
+  assertUniqueKeyBindings(
+    effectiveComponents,
+    effectiveDirectLink,
+    effectiveDirectLinkKey
+  );
```
leowla: this should not run at all if there are no direct link related changes

backend/src/db/daos/sceneDao.js
```diff
@@ -83,6 +83,48 @@ const assertDirectLinkInScenario = async (scenarioId, directLinkId) => {
   }
 };
 
+// A scene's direct link defaults to responding to Space/ArrowRight when no
+// directLinkKey is set. Mirrors directLinkKeysFor in
+// frontend/src/features/authoring/keyBindings.ts.
+const DEFAULT_DIRECT_LINK_KEYS = ["SPACE", "ARROWRIGHT"];
```
leowla: duped code bad. have one source of truth.

frontend/src/features/authoring/CanvasSideBar/SceneSettings.jsx
```diff
@@ -191,28 +245,54 @@ export default function SceneSettings() {
...
+                {keyMode === "CUSTOM" && (
+                  <div className="mt-2">
+                    <KeyCapture
+                      value={keyValue}
+                      availableKeys={availableDirectLinkKeys}
+                      onChange={saveDirectLinkKey}
+                      clearValue=""
+                    />
+                  </div>
+                )}
```
leowla, line 302: i could switch to custom key mode but never set a key and direct link is completely blocked for the player. there should be an extremely visible warning here if i have no key set

leowla (changes requested):

backend/src/db/daos/__tests__/sceneDao.test.js
kmck133: > There is a bug where the key hint doesn't show
> 
>  Screen.Recording.2026-08-22.at.5.29.07.PM.mov 
> This happens only when you copy paste components
> 
> Next Steps: Resolve on your own or make this a new ticket

erm i can't recreate this? whenever i copy paste a component it removes the key binding (intentional). could u walk me through how to get this?

kmck133: > Screen.Recording.2026-08-22.at.6.39.03.PM.mov 
> When you create these links to a slide and you delete the slides, keybinds dont get unset

should be fixed

frontend/src/features/authoring/keyBindings.ts
kmck133: " ' " triggers Firefox's "Quick Find" overlay when pressed unmodified, which would silently eat the keypress instead of firing the bound action for anyone playing in Firefox

frontend/src/features/authoring/keyBindings.ts
kmck133: its split up cause each key needs two things, how to recognize it off an event (nativeKey) and how to display it (display), like arrows show as ↑ not the word. space doesnt even have a nativeKey since its matched off e.code not e.key so it wouldnt fit a flat map anyway. NATIVE_KEY_TO_ID is just derived from NAMED_KEYS so theres still only one place to edit

backend/src/db/models/scene.js
kmck133: theyre unrelated tbh, mongoose's default just controls what gets written if the field is left out, no app meaning on its own. the actual "null means default keys" logic is in directLinkKeysFor which treats null and missing the same way anyway so it doesnt change anything either way, just a coincidence both use the word default

kmck133:
Addressed the review comments and fixed a few related bugs found while going through this.

- Key Binding is no longer nested inside Link Details — it's split out into its own collapsible section, since it doesn't just trigger the Next Scene link, it can also fire a component's State Operations with no scene change at all. Nesting it under "Link Details" made that easy to miss. The section has an info icon explaining what it actually does.
- Key Binding is now disabled unless the component has a Next Scene or a State Operation set, and it auto-clears itself if it later loses its last action (e.g. Next Scene is cleared). Previously it could be left bound with nothing to trigger.
- Deleting a scene now clears `nextScene`, `keyBinding`, and `showKeyHint` on any component elsewhere that linked to it — both in the DB and live in the editor if that scene was already open — instead of leaving a dangling link that rendered blank.
- The Direct Link toggle is disabled when the scenario only has one scene, since turning it on would have nothing to link to and would crash the player.
- Fixed `hasClickAction` treating an empty `stateOperations` array as truthy.
- `ActionableComponent`/`KeyBoundComponent` now reuse `GenericComponent` from `types.ts` instead of duplicating its shape.
- `assertUniqueKeyBindings` now skips running when a patch doesn't touch components or the direct-link fields at all.
- Moved `DEFAULT_DIRECT_LINK_KEYS`/`directLinkKeysFor` out of `sceneDao.js` into a shared `keyBindingDefaults.js` that both frontend and backend import, so they can't drift apart again.

Added tests for the new scene-deletion cleanup; all existing tests still pass.

backend/src/db/daos/sceneDao.js
```diff
@@ -4,6 +4,12 @@ import { HttpError } from "../../util/error.js";
 import status from "../../util/status.js";
 import { applyReferenceDeltas } from "./fileDao.js";
 import { HttpStatusCode } from "axios";
+// Single source of truth for direct-link key defaults lives in the
+// frontend, since it started there and the frontend has richer consumers
+// (key pickers, display strings) of the same concept. This is a plain .js
+// file specifically so this backend, which has no TypeScript loader, can
+// import it directly.
+import { directLinkKeysFor } from "../../../../frontend/src/features/authoring/keyBindingDefaults.js";
```
leowla, line 12: you'll crash prod with this because backend and frontend are deployed separately

leowla (changes requested):
Please check your AI generated code before making me review it. I'm happy to give you a full review after you test your own code.

K1mmyn:
Since leo said that please ping me on discord after you reviewed your code

kmck133:
@coderabbitai full review

PR #499, opened 2026-08-19
feat : added zoom controls to the editor

hazikchaudhry:
## Issue

There were no zoom controls before 

## Solution

Added zoom controls 
image

## Risk

might mess with layout perhaps maybe

## Checklist

- [ ] Acceptance criteria met

### DIFF
--- a/frontend/src/features/authoring/CanvasSideBar/CanvasSideBar.jsx
+++ b/frontend/src/features/authoring/CanvasSideBar/CanvasSideBar.jsx
@@ -16,7 +16,7 @@ export default function CanvasSideBar() {
   );
 
   return (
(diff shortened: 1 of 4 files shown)


K1mmyn (changes requested):
too much claude code hazik. 

there is a bug when you zoom in, you can't scroll left only right. this is because the scrollbar can only generate positive scroll values so when youre doing justify center and align items center and overflow hidden, when canvas becomes too big and you want to scroll side to side it can't because you aren't able to scoll to negative values (scroll left)

Simple fix is do justify item: safe center and align item safe center. this will prevent this scroll loss effect.

you also might want to change how much zoom is for one click. google slides does every click it double and also allows any value you want ie 69 or 420. You're basically 90% of the way there so might aswell

PR #517, opened 2026-09-01
feat: add ui feedback for prop op del

glybrio:
## Issue

No ui feedback for state operation deletion

## Solution

Added popup bottom right (like other popups).

## Risk

Hartej popping off.


harbassan (approved):
lgtm

PR #520, opened 2026-09-01
feat(authoring): show placeholder during image upload

leowla:
## Issue

Uploading an image blocks all user interaction until it is finished

## Solution

Show a non-interactive placeholder in the editor while an image is being uploaded

## Risk

None

## Checklist

- [x] Acceptance criteria met

## Summary by CodeRabbit

* **New Features**
  * Added live image-upload placeholders directly on the canvas.
  * Display upload progress with a blurred preview, progress percentage, and visual transition when complete.
  * Improved image uploads with real-time progress tracking and automatic image preloading.


harbassan (approved):
sweet, works and code looks good. One thing to point out tho is that the perecentage only over goes from 0 -> 90 -> 100, and no values in between. because of this, due you think it might be better to not have the percentage at all, and use a loading spinner instead?

leowla:
@coderabbitai review

PR #521, opened 2026-09-01
feat(authoring): centre new images at true size

leowla:


leowla:
claude auto mode getting too eager

PR #524, opened 2026-09-01
fix(authoring): layer operations on colliding zIndex

leowla:
## Issue

Layer operations like send to back do nothing when the component has the same zIndex as another.
e.g. send image backward behind textbox but both have zIndex 0.

## Solution

Renumber duplicate zIndex and do not default components to zIndex 0

## Risk

Low

## Checklist

- [x] Acceptance criteria met

## Summary by CodeRabbit

- **Bug Fixes**
  - New images and plain-text pasted text boxes now appear above existing scene elements.
  - Improved layer ordering when components share the same depth position.
  - Prevented unnecessary updates when rearranging components without changing their order.


harbassan (approved):
lgtm

PR #526, opened 2026-09-01
chore(authoring): text box writing ux

leowla:
## Issue

- Default text box font size 16px is very small
- Lorem ipsum exists to be deleted

## Solution

- 24px is bigger
- Remove default lorem ipsum
- Default to left-aligned text

## Risk

None

## Checklist

- [x] Acceptance criteria met

### DIFF
--- a/frontend/src/features/authoring/scene/operations/component.ts
+++ b/frontend/src/features/authoring/scene/operations/component.ts
@@ -16,9 +16,10 @@ export const defaults = {
     stroke: "#00000000",
     strokeWidth: 3, // default stroke width 3
     bounds: {
+      // sized to hold the placeholder text at the 32px default font size
       verts: [
         { x: 0, y: 0 },
-        { x: 400, y: 100 },
+        { x: 600, y: 100 },
       ],
       rotation: 0,
     },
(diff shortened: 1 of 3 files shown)


br-Chan (commented):
Nice, thanks. After testing it on local dev, I'm still not sure if it's big enough but it might be me. We'll see what Hartej thinks.

I was also wondering if we should get rid of the default lorem ipsum text too?

frontend/src/features/authoring/scene/operations/component.ts
br-Chan:
```suggestion
              text: "",
```

br-Chan: This is what I'd suggest. Even if it's blank, you can still find textboxes by hovering over them.

harbassan (approved):
yep it is 24 now

frontend/src/features/authoring/scene/operations/component.ts
harbassan: whats wrong with the lorem?

leowla: i have to agree, the lorem ipsum is nice for us but annoying for a real user who will have to delete it every time before typing

PR #540, opened 2026-09-05
fix: duplicate then undo crashing

harbassan:
## Issue

duplicate then undo crashing, because a render was occurring between component deletion from store and selecting clearing, therefore ui trying to render selection overlay for non-existent component.

## Solution

cleared selection immediately after operation.

## Risk

inefficient, doesn't resolve the core issue of the extra render that occurs when all post operation actions should be batched before next render.

## Checklist

- [ ] Acceptance criteria met

## Summary by CodeRabbit

* **Bug Fixes**
  * Undo and redo now correctly reset component selections while restoring editor changes, then preserve selections for restored components.


harbassan:
@coderabbitai review

